PR #8, opened 2019-05-22
Decrypt bytes

skeet70:


src/shim/ShimUtils.ts
```diff
@@ -111,19 +111,19 @@ export function validateDocumentData(data: Uint8Array) {
...
+export function validateEncryptedDocument(documentData: Base64Bytes | Uint8Array) {
     let encryptedByteLength: number;
+    if (typeof documentData !== "string" && documentData.constructor.name !== "Uint8Array") {
+        throw new Error("Invalid encrypted document content. Content should be a Base64 string or Uint8Array.");
+    }
     try {
-        //This will throw if the documentData isn't valid base64
-        encryptedByteLength = byteLength(documentData);
+        //The byteLength call will throw if the documentData isn't valid base64
+        encryptedByteLength = typeof documentData === "string" ? byteLength(documentData) : documentData.byteLength;
```
coltfred: Seems like doing a single typeof and assigning it to a local variable might make this and line 119 a little more clear.

src/shim/ShimUtils.ts
```diff
@@ -111,19 +111,19 @@ export function validateDocumentData(data: Uint8Array) {
...
     } catch (_) {
-        throw new Error("Invalid encrypted document content. Content length is not a multiple of 4 which is required for base64 bytes.");
+        throw new Error("Invalid encrypted document content. Content length is not a multiple of 4.");
```
coltfred: The only way this catch can occur is the `byteLength` function so I don't think the error message should necessarily change. Personally I'd make this more clear by doing it all inside of an if/else:
```
if(isString){
    try{...}catch()
} else {
    documentData.byteLength
}
```

src/shim/sdk/DocumentSDK.ts
```diff
@@ -112,17 +112,17 @@ export function decryptFromStore(documentID: string) {
 /**
  * Decrypt the provided document given the ID of the document and its data. Returns a Promise which will be resolved once the document has been successfully decrypted.
  * @param {string}      documentID   Unique ID of document to decrypt
- * @param {Base64Bytes} documentData Document data to decrypt
+ * @param {Base64Bytes | Uint8Array} documentData Document data to decrypt
  */
-export function decrypt(documentID: string, documentData: Base64Bytes) {
+export function decrypt(documentID: string, documentData: Base64Bytes | Uint8Array) {
```
ernieturner: We also need to update the `https://github.com/IronCoreLabs/ironweb/blob/master/ironweb.d.ts` file so the proper types are published.

src/shim/ShimUtils.ts
skeet70: Yeah, meant to change the comment back. I'm not sure what that change to doing it inside an if else clarifies. Keep in mind they aren't expressions in JS, so there'd be some more assignments in there.

ernieturner:
> Would it make sense to just have 2 different decrypt functions so the top types are more clear? decryptBase64 and decrypt?

I think I'm fine with just having a single function that takes two different types.

> Also, we should probably add a version of encrypt that returns the arrays as well

We could do this via another flag on the `DocumentCreateOptions` like `encodeResultToString` or `encodeResultAsBase64` or something that defaults to true.

src/shim/ShimUtils.ts
skeet70: TS doesn't narrow correctly then, because the argument could've been modified between when the check was done and when the call to `byteLength` happens.

src/shim/ShimUtils.ts
```diff
@@ -111,19 +111,19 @@ export function validateDocumentData(data: Uint8Array) {
...
+ * Validate that the provided encrypted document is in the proper form, which should be Base64 encoded bytes or a Uint8Array of the document content and the AES IV prepended.
+ * Therefore validate the length is at least 1 more than the IV length.
  */
-export function validateEncryptedDocument(documentData: Base64Bytes) {
-    if (typeof documentData !== "string") {
-        throw new Error(`Invalid encrypted document format provided. Expected a string.`);
-    }
+export function validateEncryptedDocument(documentData: Base64Bytes | Uint8Array) {
     let encryptedByteLength: number;
+    if (typeof documentData !== "string" && documentData.constructor.name !== "Uint8Array") {
```
ernieturner: Pretty minor, but I think it's better to do `data instanceof Uint8Array` here instead of checking `constructor.name`.

skeet70:
@coltfred @ernieturner should be ready for more review. Not sure if the change to bytes needs to go deeper than this or not, but it seems like I may have the boundaries right? 

Since this is a breaking change, do I version bump in the PR?

integration/components/documents/LocalDocument.tsx
```diff
@@ -76,7 +76,7 @@ export default class LocalList extends React.Component<LocalDocumentProps, Local
         IronWeb.document
             .updateEncryptedData(existingDoc.id, IronWeb.codec.utf8.toBytes(JSON.stringify(newList)))
             .then((encryptedDocument) => {
-                DocumentDB.updateDoc(encryptedDocument.documentID, encryptedDocument.document, encryptedDocument.documentName);
+                DocumentDB.updateDoc(encryptedDocument.documentID, IronWeb.codec.utf8.fromBytes(encryptedDocument.document), encryptedDocument.documentName);
```
ernieturner: Shouldn't this be using `codec.base64`?

integration/components/documents/NewDocument.tsx
```diff
@@ -87,7 +87,7 @@ export default class NewDocument extends React.Component<NewDocumentProps, NewDo
         IronWeb.document
             .encrypt(IronWeb.codec.utf8.toBytes(JSON.stringify(document)), documentOptions)
             .then((encryptedDocument) => {
-                DocumentDB.saveDoc(encryptedDocument.documentID, encryptedDocument.document, encryptedDocument.documentName);
+                DocumentDB.saveDoc(encryptedDocument.documentID, IronWeb.codec.utf8.fromBytes(encryptedDocument.document), encryptedDocument.documentName);
```
ernieturner: Ditto here.

PR #17, opened 2019-07-22
createNewUser that doesn't create devices

skeet70:
- added new thread through to API that creates user without creating devices
- moved old `createUser` to `createUserAndDevice`. 
- `ErrorCodes` changed to an enum to automatically have a type

TODO: 
- [x]  still need tests on a bunch of the new code path.
- [x]  remove comments and clean up, consolidate some duplication.


src/shim/Initialize.ts
```diff
@@ -58,14 +67,14 @@ function invokePasscodeCallback(passcodeCallback: PasscodeCallbackToPromise, doe
  * @param  {boolean} doesUserExist Denotes whether we need to create or update this user
  * @param  {string}  passcode      Users passcode entry
  */
-function setUserPasscode(doesUserExist: boolean, passcode: string) {
+function setUserPasscode(doesUserExist: boolean, passcode: string): Future<Error, SDKInitializationResult> {
     return getJWT(userJWTCallback)
         .flatMap((jwtToken) => {
             const payload = {
-                type: doesUserExist ? "GEN_DEVICE_KEYS" : "CREATE_USER",
```
ernieturner: Since we're re-using an existing message string to mean something different now, this will require that we major version the SDK. If we don't, a user on a prior version of the shim will get a later version of the frame and the message they pass will no longer mean what it used to and break.

src/shim/index.ts
```diff
@@ -11,6 +11,23 @@ function supportsRandomNumGen() {
     const nativeCrypto: Crypto = window.msCrypto || window.crypto;
     return typeof nativeCrypto === "object" && typeof nativeCrypto.getRandomValues === "function";
 }
+
+export const createNewUser = (jwtCallback: JWTCallbackToPromise, passcode: string): Promise<UserCreateResponse> => {
+    if (!jwtCallback || typeof jwtCallback !== "function") {
+        throw new Error("You must provide a function which will generate a JWT as the first parameter to 'IronWeb.initialize'.");
+    }
+    // TODO: find out if this is neccessary for user keygen. Assuming it is.
```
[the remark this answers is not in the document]
skeet70: Yep, removed comment.

src/shim/Initialize.ts
```diff
@@ -74,13 +83,23 @@ function setUserPasscode(doesUserExist: boolean, passcode: string) {
...
 }
 
+// Not sure if this should set the cached userJWTCallback or not, seems like it would be a one off, not repeatedly used/overriding the current user's JWT callback.
+export const createNewUser = (jwtCallback: JWTCallbackToPromise, passcode: string): Promise<UserCreateResponse> =>
+    getJWT(jwtCallback)
+        .flatMap((jwtToken) => {
+            const payload: CreateUserRequest = {type: "CREATE_USER", message: {passcode, jwtToken}};
+            return FrameMediator.sendMessage<CreateUserResponse>(payload);
+        })
+        .map(({message: {id, ...rest}}) => ({accountId: id, ...rest}))
```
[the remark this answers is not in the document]
skeet70: Done.

src/shim/Initialize.ts
skeet70: Noted, we'll have to bump major.

src/frame/initialization/tests/index.test.ts
```diff
@@ -4,13 +4,18 @@ import * as TestUtils from "../../../tests/TestUtils";
...
+import {publicKeyToBase64} from "../../../lib/Utils";
+import {fromByteArray} from "base64-js";
 
 describe("init index", () => {
-    describe("initialize", () => {
-        afterEach(() => {
-            localStorage.clear();
-        });
+    beforeEach(() => {
+        (ApiState as any).apiUser = undefined;
```
ernieturner: Should we maybe use `ApiState.clearCurrentUser()` instead of accessing the private property or did you need to only clear that one field?

skeet70: Yes, and I considered that at the time. I wasn't sure why the code was doing what it was doing in the first place and not that. I was just abstracting this up to a `beforeEach` because most of the tests either were doing this or should have been doing this.

ernieturner:
@clintfred There will be, I wrote it up [here](https://github.com/IronCoreLabs/website/issues/163) and we'll do that once all the other changes we're making are done.

Issue #12, opened 2019-06-26
Support creation of device keys without storing them locally in the browser


ernieturner:
Similar to the server side SDKs, we should have a pre-initialize (i.e. sibling of `initialize`) that lets a caller generate a new set of device keys given a valid JWT for a user and that users password. This should just return all the same information that is returned in the other SDKs (encryption/signing key pairs, internal segment ID, etc). None of this data should replace any storage of the device keys that might already exist in local storage.

```
IronWeb.generateDeviceKeys(jwt: string, password: string, {deviceName: string}?): Promise<{
  accountID: string;
  segmentID: number;
  deviceKeys: {publicKey, privateKey},
  signingKeys: {publicKey, privateKey}
}>
```

PR #18, opened 2019-07-24
#12 Add top level method to let users generate a set of device keys f…

ernieturner:
…or the provided user that are not stored in the browser and instead are returned to the caller


src/frame/initialization/index.ts
```diff
@@ -73,9 +77,9 @@ function handleVerifyResult(
  * @param {CallbackToPromise} jwtCallback                  Method which when invoked will return a Promise which will be resolved with a JWT token
  * @param {string}            deviceAndSigningSymmetricKey Optional symmetric key that is used to decrypt the users device and signing keys
  */
-export function initialize(jwtToken: string, deviceAndSigningSymmetricKey?: string) {
+export const initialize = (jwtToken: string, deviceAndSigningSymmetricKey?: string) => {
```
skeet70: Can drop the `{ return }` now too since it's a single sequence. Same comment in a few places for this file.

PR #23, opened 2019-08-19
Unmanaged decrypt

coltfred:


ironweb.d.ts
```diff
@@ -58,6 +58,9 @@ export interface DocumentMetaResponse extends DocumentAssociationResponse {
 export interface DecryptedDocumentResponse extends DocumentMetaResponse {
     data: Uint8Array;
 }
+export interface DecryptedUnmanagedDocumentResponse {
+    data: Uint8Array;
```
clintfred: Should the DocumentId and UserOrGroup that gave access be included in this response?

src/frame/sdk/DocumentAdvancedApi.ts
```diff
@@ -0,0 +1,32 @@
...
+import {documentToByteParts} from "../FrameUtils";
+import SDKError from "../../lib/SDKError";
+import {ErrorCodes} from "../../Constants";
+import EncryptedDekEndpoints from "../endpoints/EncryptedDekEndpoints";
+
+interface DecryptedUnmanagedDocumentResponse {
+    data: Uint8Array;
+}
+/**
+ * Retrieves sharing/meta information for the provided document ID. Then decrypts the provided data using the current users key and returns formatted doc structure.
```
clintfred:
```suggestion
 * Retrieves sharing/meta information for the provided document ID. Then decrypts the provided data using the current user's key and returns formatted doc structure.
```

src/frame/sdk/DocumentAdvancedApi.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function decryptLocalDoc(edeks: Uint8Array, encryptedDocument: Uint8Array): Future<SDKError, DecryptedUnmanagedDocumentResponse> {
+    //Early verification to check that the bytes we got appear to be an IronCore encrypted document. We have two versions so reject early if the bytes provided
+    //don't match either of those two versions.
+    if (encryptedDocument[0] !== 1 && encryptedDocument[0] !== 2) {
+        return Future.reject(
+            new SDKError(new Error("Provided encrypted document doesn't appear to be valid. Invalid version."), ErrorCodes.DOCUMENT_HEADER_PARSE_FAILURE)
+        );
+    }
+    const documentParts = documentToByteParts(encryptedDocument);
```
clintfred: If you decide deserializing the proto is reasonable, it be good to make sure the `documentID` in the edeks and document are the same.

src/frame/sdk/DocumentAdvancedApi.ts
```diff
@@ -0,0 +1,32 @@
...
+
+interface DecryptedUnmanagedDocumentResponse {
+    data: Uint8Array;
+}
+/**
+ * Retrieves sharing/meta information for the provided document ID. Then decrypts the provided data using the current users key and returns formatted doc structure.
+ * @param {string}     documentID        Unique lookup key of document to decrypt
+ * @param {Uint8Array} encryptedDocument Data of document to decrypt
+ */
+export function decryptLocalDoc(edeks: Uint8Array, encryptedDocument: Uint8Array): Future<SDKError, DecryptedUnmanagedDocumentResponse> {
```
clintfred: Do we want to do any validation of the edek data? I don't know what the performance characteristics are of deserializing here, but performance concerns are the main downside that I can see.

ernieturner: Pretty minor, but I'd maybe change this to be called `decryptWithProvidedEdeks` or `decryptWithEdeks` as I think that's much clearer than its current name.

ernieturner: It looks like they're getting partially validated down below via the `validateEncryptedDeks` method that gets called before we get to the frame code here.

src/shim/ShimUtils.ts
```diff
@@ -123,6 +123,19 @@ export function validateEncryptedDocument(documentData: Uint8Array) {
...
+ */
+export function validateEncryptedDeks(edeks: Uint8Array) {
+    if (!(edeks instanceof Uint8Array)) {
+        throw new Error("Invalid encrypted deks. Edeks should be a Uint8Array.");
+    }
+
+    if (edeks.byteLength <= 0) {
+        throw new Error(`Invalid encrypted deks. Length does not meet minimum requirements.`);
+    }
+}
```
ernieturner: Unit tests for this method.

ernieturner: Oh, I see you're basically testing this via the `decryptUnmanaged` calls below, so that's probably fine.

src/shim/sdk/tests/DocumentSDK.test.ts
```diff
@@ -888,4 +888,80 @@ describe("DocumentSDK", () => {
...
+                            },
+                        },
+                        [doc]
+                    );
+                    done();
+                })
+                .catch((e) => fail(e.message));
+        });
+
+        it("passes through content byte arrays", (done) => {
```
ernieturner: What is this test testing that the one above it isn't already doing?

src/shim/sdk/DocumentSDK.ts
```diff
@@ -324,3 +325,32 @@ export function revokeAccess(documentID: string, revokeList: DocumentAccessList)
...
+
+export const advanced = {
+    /**
+     * Decrypt the provided document given the edeks of the document and its data. Returns a Promise which will be resolved once the document has been successfully decrypted.
+     * @param {Uint8Array}  edeks        The encrypted deks for the documentData.
+     * @param {Uint8Array}  documentData Document data to decrypt
+     */
+    decryptUnmanaged: (edeks: Uint8Array, documentData: Uint8Array) => {
+        ShimUtils.checkSDKInitialized();
+        ShimUtils.validateEncryptedDeks(edeks);
```
skeet70: Since this can throw you probably want to catch it inside the `Future` for this function.

edit: I see the other validation functions aren't used this way, and I'm honestly not sure why. I generally don't expect a `Promise` based function to throw, and there's no `@throws` in the jsdoc that would make me think it would. @ernieturner is this intentional?

src/shim/sdk/DocumentSDK.ts
```diff
@@ -324,3 +325,32 @@ export function revokeAccess(documentID: string, revokeList: DocumentAccessList)
...
         .toPromise();
 }
+
+export const advanced = {
+    /**
+     * Decrypt the provided document given the edeks of the document and its data. Returns a Promise which will be resolved once the document has been successfully decrypted.
+     * @param {Uint8Array}  edeks        The encrypted deks for the documentData.
+     * @param {Uint8Array}  documentData Document data to decrypt
+     */
+    decryptUnmanaged: (edeks: Uint8Array, documentData: Uint8Array) => {
```
skeet70:
```suggestion
    decryptUnmanaged: (edeks: Uint8Array, documentData: Uint8Array): Promise<{data: Uint8Array; documentID: string}> => {
```

src/shim/sdk/DocumentSDK.ts
ernieturner: Sort of, ya. We communicate Promise based failures with an error code that explains what went wrong with the operation and I didn't want to create error codes for invalid data being passed in. Looking back at it I don't feel super great about how it works, but I think we've probably just got to live with it at this point.

src/frame/endpoints/EncryptedDekEndpoints.ts
```diff
@@ -0,0 +1,41 @@
...
+function getSignatureHeader() {
+    const {segmentId, id} = ApiState.user();
+    return ApiRequest.getRequestSignature(segmentId, id, ApiState.signingKeys());
+}
+
+/**
+ * Transform one of the edeks for the logged in user.
+ * @param {Uint8Array} edeks The encrypted deks to send.
+ */
+function encryptedDekTransform(protoBytes: Uint8Array) {
```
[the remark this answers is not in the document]
coltfred: I was just using a similar style to the others so I'm gonna leave it.

ironweb.d.ts
clintfred: ¯\_(ツ)_/¯ The server returns it. 

It allows the caller to understand why they can decrypt the data. It's similar to the `association` from Document GET and Document List, albeit more specific. 

We are currently returning it in IronOxide, so SDK consistency is why I'm even mentioning it in the first place.

src/frame/sdk/DocumentAdvancedApi.ts
clintfred: If this is to be delayed until encrypt is implemented, it should be added as an AC on that ticket. From a dependency perspective, I think it makes more sense to add it now, but it's not a show-stopper for me. 

Again, I'm not sure it's worth doing at all due to performance, but I suspect the performance will be acceptable.

ironweb.d.ts
```diff
@@ -58,14 +58,16 @@ export interface DocumentMetaResponse extends DocumentAssociationResponse {
...
 export interface EncryptedDocumentResponse extends DocumentIDNameResponse {
     document: Uint8Array;
 }
 export interface DocumentAccessResponse {
-    succeeded: {
-        id: string;
-        type: "user" | "group";
-    }[];
+    succeeded: UserOrGroup[];
     failed: {
```
ernieturner: Feel free to leave it like it is, but you could update this to be 

```
failed: (UserOrGroup & {error: string})[]
```

ironweb.d.ts
```diff
@@ -183,6 +188,11 @@ export class SDKError extends Error {
     rawError: Error;
 }
 
+export interface UserOrGroup {
+    type: "user" | "group";
+    id: string;
+}
```
ernieturner: Also pretty minor, but it might be nice for future us when reading this file if you moved this up above where it's used.

Issue #34, opened 2019-09-05
Tweak grantToAuthor flag to avoid major version bump


ernieturner:
PR #30 was implemented in such a way that it made a breaking change to the shim/frame communication which caused us to bump things a major version. However, the addition was very minor and shouldn't have caused consumers to have to look into why a new major version was released. We should modify the code so that it will work for earlier versions so that this can just become a 3.x version instead.

PR #35, opened 2019-09-05
#34 Only minor bump and handle case where users are on the old shim

ernieturner:


src/frame/index.ts
```diff
@@ -71,7 +71,7 @@ function onParentPortMessage(data: RequestMessage, callback: (message: ResponseM
                 data.message.documentName,
                 data.message.userGrants,
                 data.message.groupGrants,
-                data.message.grantToAuthor
+                data.message.grantToAuthor !== false
```
coltfred: We should put comments on these things. I was thinking:
```
Here we check to see if this is not false because versions < 3.1 send nothing in this spot so we have to default 
the value to ensure the types are actually correct. Note that the types >= 3.1 are actually still consistent as well, 
we just resolve the optionality in the shim.
```

ernieturner: Done

PR #36, opened 2019-09-05
Add policy grants to document create

coltfred:


src/frame/endpoints/PolicyApiEndpoints.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Generate signature message from current user state
+ */
+function getSignatureHeader() {
+    const {segmentId, id} = ApiState.user();
+    return ApiRequest.getRequestSignature(segmentId, id, ApiState.signingKeys());
+}
+
+/**
```
clintfred: This comment block doesn't make sense.

src/frame/endpoints/PolicyApiEndpoints.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Transform one of the edeks for the logged in user.
+ * @param {Uint8Array} edeks The encrypted deks to send.
+ */
+export function applyPolicy(req: Policy) {
```
clintfred: an alternative name here could be `evalPolicy`

src/frame/endpoints/PolicyApiEndpoints.ts
```diff
@@ -0,0 +1,56 @@
...
+    };
+}
+
+export default {
+    /**
+     * Call the `policies` endpoint passing all the set values in the `policyApplyRequest` as needed.
+     * @param policyApplyRequest - The policy which you want to get users and groups for
+     */
+    callApplyPolicyApi(policyApplyRequest?: Policy): Future<SDKError, PolicyApplyResponse> {
+        if (policyApplyRequest === undefined) {
```
clintfred: I'm probably just not thinking in TS here, but why even support this undefined case?

src/frame/sdk/DocumentApi.ts
```diff
@@ -74,6 +68,57 @@ function normalizeUserAndGroupPublicKeyList(userKeys: UserKeyListResponseType, g
     ];
 }
 
+/**
+ * Get a list of user and group keys for the provided users and then also add in the current user to the list of user keys
+ * @param {string[]} userGrants  List of users to get public keys for
+ * @param {string[]} groupGrants List of groups to get public keys for
+ * @param {boolean}  grantToAuthor If the logged in user should be included in the list of users to encrypt to.
+ * @param COLT:
```
clintfred: COLT!

src/frame/sdk/DocumentApi.ts
```diff
@@ -228,25 +273,25 @@ export function decryptLocalDoc(documentID: string, encryptedDocument: Uint8Arra
...
- * @param {Uint8Array} document     Document data to store
- * @param {string}     documentName Optional name of the document
- * @param {string[]}   userGrants   List of user IDs to grant access
- * @param {string[]}   groupGrants  List of group IDs to grant access
+ * @param {string}     documentID    Unique lookup key to use for document
+ * @param {Uint8Array} document      Document data to store
+ * @param {string}     documentName  Optional name of the document
+ * @param {string[]}   userGrants    List of user IDs to grant access
+ * @param {string[]}   groupGrants   List of group IDs to grant access
+ * @param {boolean}    grantToAuthor COLT: Do this
```
clintfred: TODOs

src/frame/endpoints/PolicyApiEndpoints.ts
coltfred: Sorry about that.

src/frame/endpoints/PolicyApiEndpoints.ts
coltfred: We call it policy application elsewhere, so just trying to be consistent.

src/frame/endpoints/PolicyApiEndpoints.ts
coltfred: It makes the caller not have to resolve the optionality of the request which makes places that call this cleaner. We do this in other places, like `callUserKeyListApi`.

src/frame/sdk/DocumentApi.ts
coltfred: fixed.

src/frame/sdk/DocumentApi.ts
coltfred: yep.

src/frame/endpoints/PolicyApiEndpoints.ts
```diff
@@ -0,0 +1,56 @@
...
+function getSignatureHeader() {
+    const {segmentId, id} = ApiState.user();
+    return ApiRequest.getRequestSignature(segmentId, id, ApiState.signingKeys());
+}
+
+/**
+ * Create the policy apply request based on the `policy`.
+ * @param policy The policy to be applied.
+ */
+export function applyPolicy(policy: Policy) {
```
ernieturner: I don't think this function needs to be exported, it shouldn't be used outside this file.

src/frame/endpoints/PolicyApiEndpoints.ts
```diff
@@ -0,0 +1,56 @@
...
+    const classification = policy.classification !== undefined ? [`classification=${encodeURIComponent(policy.classification)}`] : [];
+    const category = policy.category !== undefined ? [`category=${encodeURIComponent(policy.category)}`] : [];
+    const dataSubject = policy.dataSubject !== undefined ? [`dataSubject=${encodeURIComponent(policy.dataSubject)}`] : [];
+    const id = policy.substituteId !== undefined ? [`id=${encodeURIComponent(policy.substituteId)}`] : [];
```
ernieturner: It might be better to just do a more falsy check on these instead of explicitly checking `undefined`, e.g. `const category = policy.category ? ...`. We don't really want to send in a value if the user gives us an object with `null` or empty string values.

src/frame/sdk/DocumentApi.ts
```diff
@@ -16,50 +17,43 @@ import * as Utils from "../../lib/Utils";
...
-                userKeys: {result: [...userKeys.result, ...maybeCurrentUser]},
-                groupKeys,
-            });
-        }
-    );
+    groupGrants: string[]
+) {
+    const existingUserIDs = userKeys.result.map(({id}) => id);
+    const existingGroupIDs = groupKeys.result.map(({id}) => id);
+    const missingPolicyUsers = invalidPolicyUsersAndGroups.filter((userOrGroup) => userOrGroup.type === "user").map((user) => user.id);
```
ernieturner:
```suggestion
    const missingPolicyUsers = invalidPolicyUsersAndGroups.filter((userOrGroup) => userOrGroup.type === "user").map(({id}) => id);
```

src/frame/sdk/DocumentApi.ts
```diff
@@ -16,50 +17,43 @@ import * as Utils from "../../lib/Utils";
...
+    const missingPolicyGroups = invalidPolicyUsersAndGroups.filter((userOrGroup) => userOrGroup.type === "group").map((group) => group.id);
```
ernieturner:
```suggestion
    const missingPolicyGroups = invalidPolicyUsersAndGroups.filter((userOrGroup) => userOrGroup.type === "group").map(({id}) => id);
```

src/frame/sdk/DocumentApi.ts
```diff
@@ -74,6 +68,57 @@ function normalizeUserAndGroupPublicKeyList(userKeys: UserKeyListResponseType, g
...
+            );
+            return Future.reject(
+                new SDKError(
+                    new Error(
+                        `Failed to create document due to unknown users or groups in access list. Missing user IDs: [${missingUsersString}]. Missing group IDs: [${missingGroupsString}]`
+                    ),
+                    ErrorCodes.DOCUMENT_CREATE_WITH_ACCESS_FAILURE
+                )
+            );
+        } else if (userKeys.result.length === 0 && groupKeys.result.length === 0 && !grantToAuthor && policyResults.usersAndGroups.length === 0) {
```
ernieturner: This can just be an `if` since you're returning in the `if` above.

src/shim/sdk/DocumentSDK.ts
```diff
@@ -202,6 +204,7 @@ export function encrypt(documentData: Uint8Array, options?: DocumentCreateOption
             userGrants,
             groupGrants,
             grantToAuthor: encryptOptions.accessList.grantToAuthor,
+            policy: encryptOptions.policy,
```
clintfred: policy needs to be added to the docs for `encrypt` (above)

src/frame/tests/index.test.ts
```diff
@@ -231,6 +232,7 @@ describe("frame index", () => {
                     userGrants: "list of user ids",
                     groupGrants: "list of group ids",
                     grantToAuthor: true,
+                    undefined,
```
ernieturner:
```suggestion
                    policy: undefined,
```

src/frame/tests/index.test.ts
```diff
@@ -246,7 +248,8 @@ describe("frame index", () => {
                     "given name",
                     "list of user ids",
                     "list of group ids",
-                    true
+                    true,
+                    undefined
```
ernieturner: Maybe add a test (new or one of the existing ones) where you actually pass a policy?

ironweb.d.ts
```diff
@@ -15,6 +15,16 @@ export interface DocumentCreateOptions {
...
+    policy?: Policy;
+}
+/**
+ * A policy which can be used to determine the users and groups to share with.
+ */
+export interface Policy {
+    classification?: string;
+    category?: string;
+    dataSubject?: string;
+    substituteId?: string;
```
ernieturner: In IronOxide these names are different. They should probably be the same.

coltfred: ok.

src/frame/endpoints/PolicyApiEndpoints.ts
coltfred: It's used in the tests.

src/frame/endpoints/PolicyApiEndpoints.ts
coltfred: The idea of falsey makes me :nauseated_face: so I'll just take your word for it.

src/frame/sdk/DocumentApi.ts
coltfred: ok.

src/frame/tests/index.test.ts
coltfred: ok.

src/frame/sdk/DocumentApi.ts
```diff
@@ -16,50 +17,43 @@ import * as Utils from "../../lib/Utils";
...
- * Get a list of user and group keys for the provided users and then also add in the current user to the list of user keys
- * @param {string[]} userGrants  List of users to get public keys for
- * @param {string[]} groupGrants List of groups to get public keys for
- * @param {boolean}  grantToAuthor If the logged in user should be included in the list of users to encrypt to.
+ * Create the error strings for missing users/groups.
+ * @param userKeys The response from UserList
+ * @param groupKeys The response from GroupList
+ * @param invalidPolicyUsersAndGroups The invalid groups/users from policy application.
+ * @param userGrants The list of all the user grants.
+ * @param groupGrants The list of all the group grants.
```
ernieturner: No need to change, but in the future, feel free to just nuke all these `@param` lines in the comment blocks. I've usually reverted to having the doc comment have a description of what the function does, but not commenting each parameter since our penchant for good names and TS type info really gives you enough info.

coltfred: I'll delete it now since it's new.

Issue #37, opened 2019-09-12
Add timeout to frame loading to reject Promise if frame can't be loaded


ernieturner:
We recently had an outage in stage which is the environment we point to for local CMK development. The behavior of this outage meant that the IronWeb frame couldn't be loaded, but we don't have any failure checks to detect that (nor apparently is there a way to add an `onerror` handler to iframes). This meant that calls to `IronWeb.initialize` just never resolved and waited forever. 

We should add a timeout from the point that we try to append the iFrame so that we only wait a few seconds for the frame to load before we bail out and reject the Promise with an error.

Issue #28, opened 2019-08-28
Buffer is transfered to the frame on decrypt.


coltfred:
In order to avoid copying potentially large amounts of data we currently transfer ownership of the underlying buffer when we encrypt/decrypt. This works well when we have a buffer that won't be reused, but does not work well if the buffer will be reused (as it causes all Uint8Arrays on top of that buffer to appear empty).

We should not transfer the buffers across to the frame to avoid callers seeing this kind of issue.

In order to work around this, callers can call `.slice(0)` on their Uint8Arrays before passing them to cause a copy to be made.

ernieturner:
One option for a fix here would be to only do a copy to the frame when the buffer we have is only a view into the underlying buffer. That might be pretty confusing and cause odd behavior. A less performant but more complete solution would be to just copy on every clone into the frame.

PR #50, opened 2019-10-08
#28 #37 Fix a couple bugs for byte array zeroing and failure handling if the IronWeb frame can't be loaded

ernieturner:
#28 Clone byte arrays on input before passing to the frame to prevent confusing bugs by zeroing out users provided data

#37 Add timeout so operations to the frame can fail if the frame fails to load for some reason


src/shim/sdk/DocumentSDK.ts
```diff
@@ -129,7 +129,7 @@ export function decrypt(documentID: string, documentData: Uint8Array) {
         type: "DOCUMENT_DECRYPT",
         message: {
             documentID,
-            documentData,
+            documentData: documentData.slice(),
```
skeet70: https://jsperf.com/new-array-vs-splice-vs-slice/151

Looks like the right one! `slice(0)` is a bit faster for me on FF, but `slice()` is only barely slower. In Chrome `slice()` is significantly the fastest.

src/shim/sdk/DocumentSDK.ts
```diff
@@ -392,7 +392,7 @@ export const advanced = {
         const payload: MT.DocumentUnmanagedEncryptRequest = {
             type: "DOCUMENT_UNMANAGED_ENCRYPT",
             message: {
-                documentData,
+                documentData: documentData.slice(0),
```
skeet70: Why the difference here?

src/shim/sdk/tests/DocumentSDK.test.ts
```diff
@@ -246,35 +246,9 @@ describe("DocumentSDK", () => {
...
-                            message: {
-                                documentID: "mydoc",
-                                documentData: doc,
-                            },
-                        },
-                        [doc]
-                    );
+                    //Ensure we cloned these bytes before passing them to the frame
+                    const passedDoc = (FrameMediator.sendMessage as jasmine.Spy).calls.mostRecent().args[0].message.documentData;
+                    expect(passedDoc).not.toBe(doc);
```
skeet70: Probably also should check `expect(passedDoc).toEqual(doc)` if that wasn't already covered.

src/shim/sdk/tests/DocumentSDK.test.ts
```diff
@@ -319,6 +293,9 @@ describe("DocumentSDK", () => {
                         },
                         [document]
                     );
+                    //Ensure we cloned these bytes before passing them to the frame
+                    const passedDoc = (FrameMediator.sendMessage as jasmine.Spy).calls.mostRecent().args[0].message.documentData;
+                    expect(passedDoc).not.toBe(document);
```
skeet70: Same here and the rest of the spots this is being checked, unless it's covered somewhere lower down/more specific.

src/shim/sdk/DocumentSDK.ts
ernieturner: No reason, just left over from initial code. Changed.

src/shim/sdk/tests/DocumentSDK.test.ts
ernieturner: Ya, thats asserted about up above on line 239.

src/shim/sdk/tests/DocumentSDK.test.ts
ernieturner: Ditto from above, asserted about in the `.toHaveBeenCalledWith` assertion.

PR #51, opened 2019-10-16
Create user with needsRotation set

ToniaSchlick:
…n by the createNeUser function with default value false,

added needsRotation: boolean to const createNewUser,
added needsRotation to UserKey type,
added return type for getFullUser function and needsRotation to the return statment, also removed unused devices array from return statment
added needsRotation to the ApiUserResponse, UserCreateResponse and option of needsRotation to CreateUserRequest

### DIFF
--- a/ironweb.d.ts
+++ b/ironweb.d.ts
@@ -130,7 +130,7 @@ export interface UserCreateResponse {
     segmentID: number;
     status: number;
     userMasterPublicKey: PublicKey<Base64String>;
-    needsRotation?: boolean;
+    needsRotation: boolean;
 }
 export interface DeviceKeys {
     accountId: string;
@@ -273,6 +273,6 @@ export const user: User;
 export const group: Group;
 
 export function initialize(jwtCallback: JWTCallback, passcodeCallback: PasscodeCallback): Promise<SDKInitializationResult>;
-export function createNewUser(jwtCallback: JWTCallback, passcode: string): Promise<UserCreateResponse>;
+export function createNewUser(jwtCallback: JWTCallback, passcode: string, options: UserCreateOptions): Promise<UserCreateResponse>;
 export function createNewDeviceKeys(jwtCallback: JWTCallback, passcode: string): Promise<DeviceKeys>;
 export function isInitialized(): boolean;
--- a/ironweb.d.ts
+++ b/ironweb.d.ts
@@ -130,12 +130,17 @@ export interface UserCreateResponse {
     segmentID: number;
     status: number;
     userMasterPublicKey: PublicKey<Base64String>;
+    needsRotation?: boolean;
 }
 export interface DeviceKeys {
     accountId: string;
     segmentId: number;
     devicePrivateKey: PrivateKey<Base64String>;
     signingPrivateKey: PrivateKey<Base64String>;
+}
+
+export interface UserCreateOptions {
+    needsRotation?: boolean;
 }
 
 /**
--- a/ironweb.d.ts
+++ b/ironweb.d.ts
@@ -273,6 +273,6 @@ export const user: User;
 export const group: Group;
 
 export function initialize(jwtCallback: JWTCallback, passcodeCallback: PasscodeCallback): Promise<SDKInitializationResult>;
-export function createNewUser(jwtCallback: JWTCallback, passcode: string, options: UserCreateOptions): Promise<UserCreateResponse>;
+export function createNewUser(jwtCallback: JWTCallback, passcode: string, options?: UserCreateOptions): Promise<UserCreateResponse>;
 export function createNewDeviceKeys(jwtCallback: JWTCallback, passcode: string): Promise<DeviceKeys>;
 export function isInitialized(): boolean;
--- a/src/FrameMessageTypes.d.ts
+++ b/src/FrameMessageTypes.d.ts
@@ -46,7 +46,7 @@ export interface InitApiSdkResponse {
 }
 export interface CreateUserRequest {
     type: "CREATE_USER";
-    message: {passcode: string; jwtToken: string; needsRotation?: boolean};
+    message: {passcode: string; jwtToken: string; needsRotation: boolean};
 }
 export interface CreateUserResponse {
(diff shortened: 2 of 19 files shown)


ironweb.d.ts
```diff
@@ -138,6 +139,10 @@ export interface DeviceKeys {
     signingPrivateKey: PrivateKey<Base64String>;
 }
 
+export interface UserCreateOptions {
+    needsRotation?: boolean;
+}
```
ernieturner: You need to add this as the type of a new optional argument to the `createNewUser` method down below on line 276.

src/FrameMessageTypes.d.ts
```diff
@@ -46,7 +46,7 @@ export interface InitApiSdkResponse {
 }
 export interface CreateUserRequest {
     type: "CREATE_USER";
-    message: {passcode: string; jwtToken: string};
+    message: {passcode: string; jwtToken: string; needsRotation?: boolean};
```
ernieturner: For this message, the `needsRotation` boolean should be set so you should remove the `?`.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -79,11 +80,12 @@ function verify(): RequestMeta {
  * @return {RequestMeta}                Request object that can be passed to fetch() API
  */
 const userCreate = (keys: UserKeys): RequestMeta => {
```
ernieturner: Instead of adding the `needsRotation` to the `UserKeys` type (which makes the name less honest) and argument here, I'd just add a second boolean argument for `needsRotation`. That change will end up cascading down to the `callUserCreateApi` method which should take separate arguments for `keys` and `needsRotation` in this file as well as the code that calls that function.

ironweb.d.ts
```diff
@@ -130,6 +130,7 @@ export interface UserCreateResponse {
     segmentID: number;
     status: number;
     userMasterPublicKey: PublicKey<Base64String>;
+    needsRotation?: boolean;
```
ernieturner: This should always be set as a boolean, remove the `?`.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -79,11 +80,12 @@ function verify(): RequestMeta {
...
-    const {publicKey, encryptedPrivateKey} = keys;
+    const {publicKey, encryptedPrivateKey, needsRotation} = keys;
 
     const body = {
         userPublicKey: publicKeyToBase64(publicKey),
         userPrivateKey: fromByteArray(encryptedPrivateKey),
+        userNeedsRotation: needsRotation,
```
ernieturner: According to what `ironcore-id` says, the key should be `needsRotation` not `userNeedsRotation`.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -11,6 +11,7 @@ export type UserKeys = Readonly<{
     publicKey: PublicKey<Uint8Array>;
```
ernieturner: There should be unit test updates to this file to assert about the changes you're making.

ironweb.d.ts
```diff
@@ -268,6 +273,6 @@ export const user: User;
 export const group: Group;
 
 export function initialize(jwtCallback: JWTCallback, passcodeCallback: PasscodeCallback): Promise<SDKInitializationResult>;
-export function createNewUser(jwtCallback: JWTCallback, passcode: string): Promise<UserCreateResponse>;
+export function createNewUser(jwtCallback: JWTCallback, passcode: string, options: UserCreateOptions): Promise<UserCreateResponse>;
```
ernieturner: This entire `options` argument should be optional. As it is now, users would have to pass in a value for this, which we don't want them to have to do.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -78,12 +78,13 @@ function verify(): RequestMeta {
...
  * @return {RequestMeta}                Request object that can be passed to fetch() API
  */
-const userCreate = (keys: UserKeys): RequestMeta => {
+const userCreate = (keys: UserKeys, needsRotation: boolean): RequestMeta => {
     const {publicKey, encryptedPrivateKey} = keys;
 
     const body = {
         userPublicKey: publicKeyToBase64(publicKey),
         userPrivateKey: fromByteArray(encryptedPrivateKey),
+        needsRotation: needsRotation,
```
ernieturner: If the key and the variable both have the same value, you can shorthand this like

```suggestion
        needsRotation,
```

src/shim/Initialize.ts
```diff
@@ -88,15 +88,20 @@ function setUserPasscode(doesUserExist: boolean, passcode: string): Future<Error
...
+            const payload: CreateUserRequest = {type: "CREATE_USER", message: {passcode, jwtToken, needsRotation}};
             return FrameMediator.sendMessage<CreateUserResponse>(payload);
         })
         //Rename a few fields and strip out the users private key since it'll probably be confusing that they're getting back an encrypted private key
         //eslint-disable-next-line @typescript-eslint/no-unused-vars
-        .map(({message: {id, segmentId, userPrivateKey, ...rest}}) => ({accountID: id, segmentID: segmentId, ...rest}))
+        .map(({message: {id, segmentId, userPrivateKey, needsRotation, ...rest}}) => ({
+            accountID: id,
+            segmentID: segmentId,
+            needsRotation: needsRotation,
```
ernieturner: You don't have to specify the key and value if they have the same name.
```suggestion
            needsRotation,
```

ernieturner:
Also, thanks for taking the time to get all the things setup to be able to run this locally. I know it was a real pain but it'll save you a ton of time in the future and it's always good for you to know how to run our various services and how they play together. If you hadn't done it as part of this ticket you certainly would have just had to do it later.

src/shim/Initialize.ts
ToniaSchlick: done

PR #54, opened 2019-10-29
Private key rotation

ToniaSchlick:
This is not done yet. I was just hoping for some feedback.


ironweb.d.ts
```diff
@@ -146,10 +146,10 @@ export interface UserCreateOptions {
 /**
  * SDK Namespaces
  */
-
 export interface User {
     deauthorizeDevice(): Promise<{transformKeyDeleted: boolean}>;
     changePasscode(currentPasscode: string, newPasscode: string): Promise<void>;
+    rotateCurrUsersPrivateKey(passcode: string): Promise<void>;
```
ernieturner: I don't think there's any reason to put `CurrUsers` since these operations are always acting for the current use. Something like `rotateMasterKey` would be better.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "license": "AGPL-3.0-only",
-    "version": "3.1.7",
+    "version": "3.1.6",
```
ernieturner: This shouldn't need to be changed.

src/Constants.ts
```diff
@@ -46,6 +46,7 @@ export enum ErrorCodes {
     USER_DEVICE_KEY_DECRYPTION_FAILURE = 208,
     USER_PASSCODE_CHANGE_FAILURE = 209,
     USER_DEVICE_DELETE_REQUEST_FAILURE = 210,
+    USER_PRIVATE_KEY_ROTATION_FAILURE = 211,
```
ernieturner: This also needs to be added to the `ErrorCodes` enum in the `ironweb.d.ts` file.

src/WorkerMessageTypes.d.ts
```diff
@@ -72,6 +72,22 @@ export interface DecryptLocalKeysWorkerResponse {
     };
 }
 
+export interface RotateUserPrivateKeyWorkRequest {
```
ernieturner:
```suggestion
export interface RotateUserPrivateKeyWorkerRequest {
```

src/WorkerMessageTypes.d.ts
```diff
@@ -72,6 +72,22 @@ export interface DecryptLocalKeysWorkerResponse {
...
+    type: "ROTATE_USER_PRIVATE_KEY";
+    message: {
+        passcode: string;
+        privateKey: Uint8Array;
+    };
+}
+
+export interface RotateUserPrivateKeyWorkResponse {
```
ernieturner:
```suggestion
export interface RotateUserPrivateKeyWorkerResponse {
```

src/WorkerMessageTypes.d.ts
```diff
@@ -72,6 +72,22 @@ export interface DecryptLocalKeysWorkerResponse {
     };
 }
 
+export interface RotateUserPrivateKeyWorkRequest {
+    type: "ROTATE_USER_PRIVATE_KEY";
+    message: {
+        passcode: string;
+        privateKey: Uint8Array;
```
ernieturner: This should probably be named `encryptedPrivateUserKey` to be more clear about what it represents.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -179,6 +179,26 @@ const userUpdate = (userID: string, userPrivateKey?: PrivateKey<Uint8Array>, sta
...
+): RequestMeta => ({
+    url: `users/${encodeURIComponent(userID)}keys/${userKeyId}`,
+    options: {
+        method: "PUT",
+        headers: {
+            "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+            userPrivateKey: userPrivateKey ? fromByteArray(userPrivateKey) : undefined,
+            augmentationFactor: augmentationFactor ? fromByteArray(augmentationFactor) : undefined,
```
ernieturner: Can both of these really be optional?

src/frame/sdk/UserApi.ts
```diff
@@ -25,6 +25,27 @@ export function deauthorizeDevice() {
...
 
+/**
+ * Rotate users current private key by taking their current passcode and using it to derive a key to decrypt their user private key.
+ * Then generates and augmentation factor and subtracts that augmentation factor from the users private key. The new private key is then
+ * encrypted. The new encrypted private key and the augmetnation factor is then passed as an authorized api request. The server then
+ * validates the request and augments the server side private key.
+ * @param {string} passcode Users current passcode
+ */
+export function rotateCurrUsersPrivateKey(passcode: string) {
+    const {privateKey} = ApiState.signingKeys();
```
ernieturner: We're not rotating their private signing key right? I assume this should be `ApiState.encryptedUserKey()`

src/frame/worker/UserCrypto.ts
```diff
@@ -4,6 +4,25 @@ import SDKError from "../../lib/SDKError";
...
 import loadRecrypt from "./crypto/recrypt";
 
+/**
+ * Requests augmentation on users current private key. The augmented private key is then encrypted and using the password derived key.
+ * @param {Uint8Array} privateKey
+ * @param {Uint8Array} augmentationFactor
+ */
+export function augmentPrivateKey(
+    passcode: string,
+    userPrivateKey: Uint8Array
```
ernieturner: Given my comment above, I'm pretty sure what we get here is actually the users encrypted private key. We need to use the password they give us to derive their private key escrow key and decrypt the key.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -14,6 +14,22 @@ const {SALT_LENGTH, PBKDF2_ITERATIONS} = CryptoConstants;
 type Plaintext = Uint8Array;
 let RecryptApi: Recrypt.Api256;
 
+/**
+ * Augments UserPrivateKey and return both the augmented private key and the augmentation factor.
+ * @param userPrivateKey current private key to be augmented
+ * @param augmentatioFactor
+ */
+export const getRotationResults = (userPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
```
ernieturner: This function name doesn't feel very clear about what it's actually doing. Maybe just `rotateUsersPrivateKey` or something instead? Maybe there's already a name used in ironoxide that we could copy.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -14,6 +14,22 @@ const {SALT_LENGTH, PBKDF2_ITERATIONS} = CryptoConstants;
...
+    return generateKeyPair().map((data) => {
+        const augmentationFactor = data.privateKey;
+        const newPrivateKey = Recrypt.subtractPrivateKeys(userPrivateKey, augmentationFactor);
+        return {
+            newPrivateKey,
+            augmentationFactor,
+        };
```
ernieturner: There's really no need for these local variables, I'd just do

```
return {
  augmentationFactor: data.privateKey,
  newPrivateKey: Recrypt.subtractPrivateKeys(userPrivateKey, augmentationFactor)
}
```

src/frame/worker/index.ts
```diff
@@ -93,6 +93,10 @@ messenger.onMessage((data: RequestMessage, callback: (message: ResponseMessage,
                 data.message.symmetricKey,
                 data.message.nonce
             ).engage(errorHandler, (deviceAndSigningKeys) => callback({type: "DECRYPT_LOCAL_KEYS_RESPONSE", message: deviceAndSigningKeys}));
+        case "ROTATE_USER_PRIVATE_KEY":
+            return UserCrypto.augmentPrivateKey(data.message.passcode, data.message.privateKey).engage(errorHandler, (userRotationResult) =>
```
ernieturner: We should probably standardize on either "rotate" or "augment" in our naming here.

globals.d.ts
```diff
@@ -30,6 +30,7 @@ declare const _WORKER_PATH_LOCATION_: string;
 type Base64String = string;
 type PrivateKey<T> = T;
 type SigningPublicKey<T> = T;
+type AugmentationFactor<T> = T;
```
ernieturner: Does this need to be generic? Are there times when we represent this type as anything other than a `Uint8Array`?

ironweb.d.ts
```diff
@@ -272,7 +273,7 @@ export const codec: Codec;
 export const user: User;
 export const group: Group;
 
-export function initialize(jwtCallback: JWTCallback, passcodeCallback: PasscodeCallback): Promise<SDKInitializationResult>;
+export function initialize(jwtCallback: JWTCallback, passcodeCallback: PasscodeCallback, needsRotation: boolean): Promise<SDKInitializationResult>;
```
ernieturner: I don't recall what we decided to do with automatically rotating the users private key if it was marked as such after initialize. @clintfred do you recall? If we make it automatic then there's no need for this parameter. If we're making the caller manually invoke rotation then we have to make this new argument optional (with a default of false) in order to not break backward compatibility.

clintfred: We are not automatically doing any key rotation. We do want to be able _notify_ that there are private keys that are marked for rotation at initialization time and then make it reasonably easy to do those rotations. In ironoxide there are two initialize calls, one that will do the rotation checks and one that will not. https://github.com/IronCoreLabs/ironoxide/blob/9f5c7e35283803ae77d5b67b0d07c8d3111a46ee/src/lib.rs#L159

Actually doing the rotation is going to require some sort of call. I think it may be wise to eventually provide a `rotateAll` type of call that will rotate all the keys (both user and group) that are marked for rotation, but it's not strictly necessary of course.

ernieturner: Ok, if that's the case then we should remove this `needsRotation` argument here. As long as the value is exposed back to users on the post `initialize` call they can then call `SDK.user.rotatePrivateKey` manually to do the rotation on an as-needed basis.

src/frame/worker/UserCrypto.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: I did change this but not to your exact suggestion. The augmentationFactor is provided by generateKeyPair() as a privateKey.

src/frame/worker/index.ts
ToniaSchlick: went with rotate, (less to change)

src/frame/endpoints/UserApiEndpoints.ts
ernieturner: They might be optional for ironcore-id, but here both the `userPrivateKey` and `augmentationFactor` are marked as fixed fields, they aren't options. So there's no need to check whether they exist.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ernieturner: I was slightly wrong in my code sample since the `augmentationFactor` variable doesn't exist. Here's what I was trying to say this block of code can be.

```
return generateKeyPair().map(({privateKey) => ({
	augmentationFactor: privateKey,
    newPrivateKey: Recrypt.subtractPrivateKeys(userPrivateKey, privateKey)
}));
```

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -179,6 +179,21 @@ const userUpdate = (userID: string, userPrivateKey?: PrivateKey<Uint8Array>, sta
     errorCode: ErrorCodes.USER_UPDATE_REQUEST_FAILURE,
 });
 
+const userKeyUpdateApi = (userID: string, userPrivateKey: PrivateKey<Uint8Array>, augmentationFactor: AugmentationFactor, userKeyId: number): RequestMeta => ({
```
ernieturner: We should add a docblock above this method about what it does. You don't need to do any of the `@param` stuff but a sentence or two about what this is doing would be good.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -179,6 +179,21 @@ const userUpdate = (userID: string, userPrivateKey?: PrivateKey<Uint8Array>, sta
...
+        method: "PUT",
+        headers: {
+            "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+            userPrivateKey,
+            augmentationFactor,
+        }),
+    },
+    errorCode: ErrorCodes.USER_UPDATE_REQUEST_FAILURE,
```
ernieturner: I think it would be a good idea to add a unique error code for this since it's a different URL we're hitting than normal use update. So add something like `USER_UPDATE_KEY_REQUEST_FAILURE`.

src/frame/endpoints/UserApiEndpoints.ts
```diff
@@ -179,6 +179,21 @@ const userUpdate = (userID: string, userPrivateKey?: PrivateKey<Uint8Array>, sta
...
+const userKeyUpdateApi = (userID: string, userPrivateKey: PrivateKey<Uint8Array>, augmentationFactor: AugmentationFactor, userKeyId: number): RequestMeta => ({
+    url: `users/${encodeURIComponent(userID)}/keys/${userKeyId}`,
+    options: {
+        method: "PUT",
+        headers: {
+            "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+            userPrivateKey,
+            augmentationFactor,
```
ernieturner: You need to convert these byte arrays into base64 before you JSON stringify them. The `fromByteArray` method that is already imported in this file is what you'll want to use.

src/frame/endpoints/tests/UserApiEndpoints.test.ts
```diff
@@ -184,6 +184,36 @@ describe("UserApiEndpoints", () => {
...
+                (userKeys: any) => {
+                    expect(userKeys).toEqual({
+                        id: "user-10",
+                        foo: "bar",
+                    });
+                    expect(ApiRequest.makeAuthorizedApiRequest).toHaveBeenCalledWith("users/user-10/keys/1", expect.any(Number), expect.any(Object)); // keyId?
+                    const request = (ApiRequest.makeAuthorizedApiRequest as jasmine.Spy).calls.argsFor(0)[2];
+                    expect(JSON.parse(request.body)).toEqual({
+                        userPrivateKey: expect.any(Object),
+                        augmentationFactor: expect.any(Object),
```
ernieturner: You should be able to assert directly about these values since you provided them up above on the `callUserKeyUpdateApi`. After you address my comments above these values will be base64 encoded strings.

src/frame/sdk/UserApi.ts
```diff
@@ -25,6 +25,28 @@ export function deauthorizeDevice() {
...
+ * @param {string} passcode Users current passcode
+ */
+export function rotateUserMasterKey(passcode: string) {
+    const encryptedPrivateUserKey = ApiState.encryptedUserKey();
+    const payload: WMT.RotateUserPrivateKeyWorkerRequest = {
+        type: "ROTATE_USER_PRIVATE_KEY",
+        message: {
+            passcode,
+            encryptedPrivateUserKey,
+            keySalt: sliceArrayBuffer(encryptedPrivateUserKey, 0, CryptoConstants.SALT_LENGTH),
```
ernieturner: I'm not sure there's any reason to slice off the keySalt here since what we're slicing here is already full sent in the `encryptedPrivateUserKey`. The code within the worker should just do the work to slice off the salt.

src/frame/worker/UserCrypto.ts
```diff
@@ -43,6 +43,34 @@ function generateTransformKeyAndEncryptUserKey(userKeySet: KeyPairSet, derivedKe
...
+ * @param {Uint8Array} augmentationFactor
+ */
+export function rotatePrivateKey(
+    passcode: string,
+    encryptedPrivateUserKey: Uint8Array,
+    derivedKeySalt: Uint8Array
+): Future<SDKError, {newEncryptedPrivateUserKey: Uint8Array; augmentationFactor: Uint8Array}> {
+    return loadRecrypt()
+        .flatMap((Recrypt) => {
+            return Future.gather2(decryptUserMasterPrivateKey(passcode, encryptedPrivateUserKey, derivedKeySalt), Recrypt.generatePasswordDerivedKey(passcode))
```
ernieturner: Because the `decryptUserMasterPrivateKey` doesn't require `Recrypt` to work, we should do this work at the same time we do the `loadRecrypt` call. Take a look at the `changeUsersPasscode` function where we initially do a `gather2` like `return Future.gather2(decryptUserMasterPrivateKey(passcode, derivedKeySalt, encryptedPrivateUserKey), loadRecrypt())`. 

Also, you're passing the arguments here in the incorrect order, the `decryptUserMasterPrivateKey` method takes `passcode`, `salt`, and then `encryptedKey`.

ernieturner: I also have some other improvements to make here that I had briefly mentioned yesterday but we should get these changes done first before we address those.

src/frame/worker/UserCrypto.ts
```diff
@@ -43,6 +43,34 @@ function generateTransformKeyAndEncryptUserKey(userKeySet: KeyPairSet, derivedKe
...
+                .flatMap(([userPrivateKey, derivedKey]) => {
+                    return Future.gather2(Recrypt.rotateUsersPrivateKey(userPrivateKey), Future.of(derivedKey));
+                })
+                .flatMap(([{newPrivateKey, augmentationFactor}, derivedKey]) => {
+                    return AES.encryptUserKey(newPrivateKey, derivedKey).map((encryptedPrivateKey) => {
+                        return {
+                            newEncryptedPrivateUserKey: encryptedPrivateKey,
+                            augmentationFactor,
+                        };
```
ernieturner: Minor, but you can avoid the `return` in this section by doing

```
return AES.encryptUserKey(newPrivateKey, derivedKey).map((encryptedPrivateKey) => ({
    newEncryptedPrivateUserKey: encryptedPrivateKey,
    augmentationFactor,
});
```

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -95,6 +95,20 @@ export const generateSigningKeyPair = (): Future<Error, SigningKeyPair> => Futur
 export const getPublicSigningKeyFromPrivate = (privateSigningKey: PrivateKey<Uint8Array>) =>
     Future.tryF(() => RecryptApi.computeEd25519PublicKey(privateSigningKey));
 
+/**
+ * Augments UserPrivateKey and return both the augmented private key and the augmentation factor.
+ * @param userPrivateKey current private key to be augmented
+ * @param augmentatioFactor
```
ernieturner:
```suggestion

```

src/shim/sdk/UserSDK.ts
```diff
@@ -18,6 +18,20 @@ export function changePasscode(currentPasscode: string, newPasscode: string) {
         .map(() => undefined)
         .toPromise();
 }
+/**
+ * Rotates the user current private key.
+ * @param {string} passcode
```
ernieturner: You should add a comment about what this parameter is

src/frame/worker/UserCrypto.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: done

src/shim/sdk/UserSDK.ts
ToniaSchlick: done

src/frame/sdk/UserApi.ts
ToniaSchlick: done

src/frame/worker/UserCrypto.ts
ToniaSchlick: fixed the two issues.

src/frame/worker/UserCrypto.ts
```diff
@@ -163,7 +190,8 @@ export function decryptDeviceAndSigningKeys(encryptedDeviceKey: Uint8Array, encr
  * @param {Uint8Array} derivedKeySalt          Salt that was used during prior key derivation
```
ernieturner: Minor, but remove this line.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -95,6 +95,19 @@ export const generateSigningKeyPair = (): Future<Error, SigningKeyPair> => Futur
...
+ * Augments UserPrivateKey and return both the augmented private key and the augmentation factor.
+ * @param userPrivateKey current private key to be augmented
+ */
+export const rotateUsersPrivateKey = (userPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
+    return generateKeyPair().map(({privateKey}) => {
+        return {
+            newPrivateKey: Recrypt.subtractPrivateKeys(userPrivateKey, privateKey),
+            augmentationFactor: privateKey,
+        };
+    });
```
ernieturner: Very nitpicking, but you can avoid the second return by doing

```
    return generateKeyPair().map(({privateKey}) => ({
        newPrivateKey: Recrypt.subtractPrivateKeys(userPrivateKey, privateKey),
        augmentationFactor: privateKey,
    }));
```

src/frame/worker/UserCrypto.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: done

src/frame/sdk/tests/UserApi.test.ts
```diff
@@ -39,6 +39,36 @@ describe("UserApi", () => {
...
+    describe("rotateUserMasterKey", () => {
+        it("sends message to worker and takes result and passes it to call user key update endpoint", (done) => {
+            ApiState.setCurrentUser(TestUtils.getFullUser());
+            spyOn(UserApiEndpoints, "callUserKeyUpdateApi").and.returnValue(Future.of("user key update result"));
+            spyOn(WorkerMediator, "sendMessage").and.returnValue(
+                Future.of({
+                    message: {newEncryptedPrivateUserKey: "newEncryptedPrivateUserKey", augmentationFactor: "augmentationFactor"},
+                })
+            );
+            spyOn(ApiState, "setEncryptedPrivateUserKey");
```
ernieturner: No need to change this, but because this call doesn't have any side effects that we don't want (like making requests to the API or similar) mocking it out isn't really necessary. Instead of doing a mock and then asserting that it was called below, you could just not mock things and assert that the result of `ApiState.encryptedUserKey` is what you'd expect. The goal here is mostly to avoid mocking out everything and only mock out the parts that will cause a bunch of problems when testing.

__mocks__/@ironcorelabs/recrypt-wasm-binding.js
```diff
@@ -3,6 +3,10 @@
  * replace it with this mock which just returns empty values for most things. This mock is automatically consumed by Jest given its location
  */
 exports.Api256 = class Api256 {
+    subtractPrivateKeys() {
+        return new Uint8Array(32);
+    }
```
ernieturner: Remove this from here since it's defined where it should be below.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -82,7 +82,7 @@ export const instantiateApi = (seed?: Uint8Array) => {
 /**
  * Generate a new Recrypt public/private key pair
  */
-export const generateKeyPair = (): Future<Error, KeyPair> => Future.tryF(() => RecryptApi.generateKeyPair());
+export const generateKeyPair = (): Future<Error, KeyPair> => Future.tryF(() => RecryptApi.generateKeyPair()); //add logs to this to see what is being vc
```
ernieturner: Remove the comment.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -95,6 +95,34 @@ export const generateSigningKeyPair = (): Future<Error, SigningKeyPair> => Futur
 export const getPublicSigningKeyFromPrivate = (privateSigningKey: PrivateKey<Uint8Array>) =>
     Future.tryF(() => RecryptApi.computeEd25519PublicKey(privateSigningKey));
 
+const isBufferAllZero = (bytes: Uint8Array) => {
+    return bytes.every((val) => val === 0);
+};
```
ernieturner: Add a quick comment block explaining what this does. Also you could shorten this by just doing

```
const isBufferAllZero = (bytes: Uint8Array) => bytes.every((val) => val === 0);
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
 import * as CryptoUtils from "../../CryptoUtils";
 import * as nativePBKDF2 from "../../pbkdf2/native";
 import * as Recrypt from "../RecryptWasm";
+import * as MochRecrypt from "@ironcorelabs/recrypt-wasm-binding";
```
ernieturner:
```suggestion
import * as MockRecrypt from "@ironcorelabs/recrypt-wasm-binding";
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
 
+    describe("rotateUsersPrivateKeyWithRetry", () => {
+        const userPrivateKey = new Uint8Array([12, 13, 14]);
+        it("should result in an error when mocked generateKeyPair returns Uint8Array of zeros for augmentationFactor", () => {
+            jest.spyOn(Recrypt.getApi(), "generateKeyPair");
+            Recrypt.rotateUsersPrivateKeyWithRetry(userPrivateKey).engage(
+                (error) => {
+                    expect(error.message).toEqual("Key rotation failed.");
+                },
+                () => fail("Should not invoke success when operation fails")
```
ernieturner:
```suggestion
                () => fail("Should not succeed when augmentation factor is all zeros")
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+            );
+            expect(Recrypt.getApi().generateKeyPair).toHaveBeenCalledTimes(2);
```
ernieturner: This assertion should be moved into the reject handler up above on line 20.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+            spyOn(Recrypt.getApi(), "generateKeyPair").and.returnValue({privateKey: new Uint8Array([12, 23, 34])});
+            jest.spyOn(MochRecrypt, "subtractPrivateKeys").mockReturnValue(new Uint8Array(32));
+            Recrypt.rotateUsersPrivateKeyWithRetry(userPrivateKey).engage(
+                (error) => {
+                    expect(error.message).toEqual("Key rotation failed.");
+                },
+                () => fail("Should not invoke success when operation fails")
+            );
+
+            expect(MochRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(2);
```
ernieturner: Ditto about moving this up to the error handler on line 32.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+        });
+
+        it("should result in an error when subtraction result is zero", () => {
+            spyOn(Recrypt.getApi(), "generateKeyPair").and.returnValue({privateKey: new Uint8Array([12, 23, 34])});
+            jest.spyOn(MochRecrypt, "subtractPrivateKeys").mockReturnValue(new Uint8Array(32));
+            Recrypt.rotateUsersPrivateKeyWithRetry(userPrivateKey).engage(
+                (error) => {
+                    expect(error.message).toEqual("Key rotation failed.");
+                },
+                () => fail("Should not invoke success when operation fails")
```
ernieturner:
```suggestion
                () => fail("Should fail when private key subtraction results in zeroed private key")
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,40 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+            );
+
+            expect(MochRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(2);
+        });
+    });
```
ernieturner: We should also add a successful test here that returns valid values for the add and subtract and ensure that both are only called once.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -94,6 +94,37 @@ export const generateSigningKeyPair = (): Future<Error, SigningKeyPair> => Futur
...
+            return Future.reject(new Error("Key rotation failed."));
+        }
+        return Future.of({
+            newPrivateKey,
+            augmentationFactor: privateKey,
+        });
+    });
+};
+
+export const rotateUsersPrivateKeyWithRetry = (userPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
```
ernieturner: Add a docblock for this method too that explains why we're doing a retry.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,54 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+        });
+
+        it("should success when valid augmentation factor is produced and the result of the subtraction does not result in zero", () => {
+            spyOn(Recrypt.getApi(), "generateKeyPair").and.returnValue({privateKey: new Uint8Array([12, 23, 34])});
+            jest.spyOn(MockRecrypt, "subtractPrivateKeys");
+            Recrypt.rotateUsersPrivateKeyWithRetry(userPrivateKey).engage(
+                (e) => e,
+                ({newPrivateKey, augmentationFactor}) => {
+                    expect(Recrypt.getApi().generateKeyPair).toHaveBeenCalled;
+                    expect(MockRecrypt.subtractPrivateKeys).toHaveBeenCalled;
```
ernieturner:
```suggestion
                    expect(MockRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(1);
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,54 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+            expect(MockRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(2);
+        });
+
+        it("should success when valid augmentation factor is produced and the result of the subtraction does not result in zero", () => {
+            spyOn(Recrypt.getApi(), "generateKeyPair").and.returnValue({privateKey: new Uint8Array([12, 23, 34])});
+            jest.spyOn(MockRecrypt, "subtractPrivateKeys");
+            Recrypt.rotateUsersPrivateKeyWithRetry(userPrivateKey).engage(
+                (e) => e,
+                ({newPrivateKey, augmentationFactor}) => {
+                    expect(Recrypt.getApi().generateKeyPair).toHaveBeenCalled;
```
ernieturner:
```suggestion
                    expect(Recrypt.getApi().generateKeyPair).toHaveBeenCalledTimes(1);
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -4,12 +4,54 @@ import * as TestUtils from "../../../../../tests/TestUtils";
...
+                },
+                () => fail("Should fail when private key subtraction results in zeroed private key")
+            );
+
+            expect(MockRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(2);
+        });
+
+        it("should success when valid augmentation factor is produced and the result of the subtraction does not result in zero", () => {
+            spyOn(Recrypt.getApi(), "generateKeyPair").and.returnValue({privateKey: new Uint8Array([12, 23, 34])});
+            jest.spyOn(MockRecrypt, "subtractPrivateKeys");
```
ernieturner: Don't we need to forcefully return a different value here to avoid having it run into the failure case of getting back all zeros? I'm a little confused as to why this test is passing...can you dig into that?

ToniaSchlick: fixed it

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

PR #56, opened 2019-11-04
Adding Schnorr signature to request when adding members to group

ToniaSchlick:
- added signature to the GroupAddMemberWorkerResponse worker message type,
	- added signature to the addMembers Endpoint,
	- changed the name of generateGroupTransformKeyToList to generateGroupTransformKeyToListAndSignature
	- added groupPublicKey and groupID to GroupAddMemberWorkerRequest
	- created a generateAddMemberSignature function in RecryptWasm

### DIFF
--- a/src/frame/worker/GroupCrypto.ts
+++ b/src/frame/worker/GroupCrypto.ts
@@ -57,16 +57,24 @@ export function addAdminsToGroup(
  * @param {SigningKeyPair}              signingKeys     Current users signing keys used to sign transform key
  */
 export function addMembersToGroup(
-    groupPrivateKey: TransformedEncryptedMessage,
+    encryptedGroupPrivateKey: TransformedEncryptedMessage,
+    groupPublicKey: PublicKey<string>,
+    groupID: string,
     userKeyList: UserOrGroupPublicKey[],
     adminPrivateKey: PrivateKey<Uint8Array>,
     signingKeys: SigningKeyPair
 ) {
     return loadRecrypt()
         .flatMap((Recrypt) => {
-            return Recrypt.decryptPlaintext(groupPrivateKey, adminPrivateKey).flatMap(([_, key]) =>
-                Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys)
-            );
+            return Recrypt.decryptPlaintext(encryptedGroupPrivateKey, adminPrivateKey).flatMap(([groupPrivateKey, key]) => {
+                return Future.gather2(
+                    Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys),
+                    Recrypt.generateAddMemberSignature(groupPrivateKey, groupPublicKey, groupID)
+                ).map(([transformKeyGrant, signature]) => ({
+                    transformKeyGrant,
+                    signature,
+                }));
+            });
         })
         .errorMap((error) => new SDKError(error, ErrorCodes.GROUP_MEMBER_KEY_ENCRYPTION_FAILURE));
 }
--- a/src/frame/worker/GroupCrypto.ts
+++ b/src/frame/worker/GroupCrypto.ts
@@ -70,12 +70,9 @@ export function addMembersToGroup(
     return loadRecrypt()
         .flatMap((Recrypt) => {
             return Recrypt.decryptPlaintext(encryptedGroupPrivateKey, adminPrivateKey).flatMap(([_, key]) => {
-                return Future.gather2(
-                    Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys),
-                    Future.of(Recrypt.schnorrSignUtf8String(key, publicKeyToBytes(groupPublicKey), groupID))
-                ).map(([transformKeyGrant, signature]) => ({
+                return Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys).map((transformKeyGrant) => ({
                     transformKeyGrant,
-                    signature,
+                    signature: Recrypt.schnorrSignUtf8String(key, publicKeyToBytes(groupPublicKey), groupID),
                 }));
             });
         })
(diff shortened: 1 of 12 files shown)


src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,10 @@ export const createRequestSignature = (
     };
 };
 
+export const generateAddMemberSignature = (groupPrivateKey: Uint8Array, groupPublicKey: PublicKey<string>, groupId: string) => {
```
ernieturner: Add a quick docblock to this function.

ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,16 @@ export const createRequestSignature = (
     };
 };
 
+/**
+ * Generate a schnorr signature using the group's private key over the group's provided id
+ * @param groupPrivateKey
+ * @param groupPublicKey
+ * @param groupId
+ */
+export const generateAddMemberSignature = (groupPrivateKey: Uint8Array, groupPublicKey: PublicKey<string>, groupId: string) => {
```
ernieturner: Won't this same signature calculation be used when we add admins to the group too? Should this be renamed to `calculateGroupModificationSignature` or similar instead?

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,16 @@ export const createRequestSignature = (
...
+    return Future.of(RecryptApi.schnorrSign(groupPrivateKey, publicKeyToBytes(groupPublicKey), utf8StringToArrayBuffer(groupId)));
```
ernieturner: I'm not sure there's a reason for this to return a Future if it doesn't need to. Similar to the `createRequestSignature` above since there's no async calls happening I'd probably just directly return the `schnorrSign` call and wrap call so this function in a Future where needed.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,16 @@ export const createRequestSignature = (
     };
 };
 
+/**
+ * Generate a schnorr signature using the group's private key over the group's provided id
+ * @param groupPrivateKey
+ * @param groupPublicKey
+ * @param groupId
```
ernieturner: If we're not going to comment on what these params are (which I don't think is really necessary) they should just be removed from the doc block. In this form they don't help anybody.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
BobWall23: Seems like this is pretty general purpose and could maybe be called `signString` - take a private key, a public key, and a string.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: I went with schnorrSignString.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,13 @@ export const createRequestSignature = (
     };
 };
 
+/**
+ * Generate a schnorr signature using the group's private key over the group's provided id
```
ernieturner: Change this comment to be more generic about signing the provided UTF-8 string with the provided public and private signing keys.

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -248,6 +248,13 @@ export const createRequestSignature = (
...
+ */
+export const schnorrSignUtf8String = (privateKey: Uint8Array, publicKey: PublicKey<string>, string: string) => {
```
ernieturner: Sorry I didn't notice this before, but we should make this take a `PublicKey<Uint8Array>` instead of as a string. It's inconsistent that we take the private key as bytes but the public key as a string. Callers should just convert it from a string to bytes before calling into here.

ToniaSchlick: done

src/frame/worker/GroupCrypto.ts
```diff
@@ -50,23 +51,33 @@ export function addAdminsToGroup(
...
 ) {
     return loadRecrypt()
         .flatMap((Recrypt) => {
-            return Recrypt.decryptPlaintext(groupPrivateKey, adminPrivateKey).flatMap(([_, key]) =>
-                Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys)
-            );
+            return Recrypt.decryptPlaintext(encryptedGroupPrivateKey, adminPrivateKey).flatMap(([_, key]) => {
+                return Future.gather2(
+                    Recrypt.generateTransformKeyToList(key, userKeyList, signingKeys),
+                    Future.of(Recrypt.schnorrSignUtf8String(key, publicKeyToBytes(groupPublicKey), groupID))
```
ernieturner: Similar to the other PR that Colt commented on, we should change this call here to do just move it down to the `map` below.

PR #58, opened 2019-11-07
Add schnorr signature admins

ToniaSchlick:


src/frame/endpoints/tests/GroupApiEndpoints.test.ts
```diff
@@ -280,7 +281,7 @@ describe("GroupApiEndpoints", () => {
                 },
             ];
 
-            GroupApiEndpoints.callAddAdminsApi("22", userKeys).engage(
+            GroupApiEndpoints.callAddAdminsApi("22", userKeys, signature).engage(
```
ernieturner: You should add the value of this `signature` into the body assertion below.

src/frame/endpoints/GroupApiEndpoints.ts
```diff
@@ -142,6 +143,7 @@ function addAdmins(groupID: string, addedAdmins: EncryptedAccessKey[]) {
                     ...admin.encryptedPlaintext,
                 })),
             }),
+            sugnature: fromByteArray(signature),
```
coltfred: sugnature? Also you should manually test that this request will fail if the signature is invalid. This is currently being sent to the server as an invalid name (thus all the signing work you did was completely being ignored by ironcore-id).

I don't know if you can actually write a test on this? I'm pretty sure you can. @ernieturner would be more helpful on that than I will, but if he's not around I'm happy to help.

src/frame/endpoints/GroupApiEndpoints.ts
```diff
@@ -299,8 +301,8 @@ export default {
      * @param {string}               groupID   ID of group to add admins to
      * @param {EncryptedAccessKey[]} adminList List of admin users to add to the group
      */
-    callAddAdminsApi(groupID: string, adminList: EncryptedAccessKey[]) {
-        const {url, options, errorCode} = addAdmins(groupID, adminList);
+    callAddAdminsApi(groupID: string, adminList: EncryptedAccessKey[], signasture: Uint8Array) {
```
coltfred: spelling.

src/frame/worker/GroupCrypto.ts
```diff
@@ -28,23 +28,33 @@ export function createGroup(userPublicKey: PublicKey<Uint8Array>, signingKeys: S
...
 ) {
     return loadRecrypt()
         .flatMap((Recrypt) => {
-            return Recrypt.decryptPlaintext(encryptedGroupPrivateKey, adminPrivateKey).flatMap(([plaintext]) =>
-                Recrypt.encryptPlaintextToList(plaintext, userKeyList, signingKeys)
-            );
+            return Recrypt.decryptPlaintext(encryptedGroupPrivateKey, adminPrivateKey).flatMap(([plaintext, key]) => {
+                return Future.gather2(
+                    Recrypt.encryptPlaintextToList(plaintext, userKeyList, signingKeys),
+                    Future.of(Recrypt.schnorrSignUtf8String(key, groupPublicKey, groupID))
```
coltfred: I'm a bit confused why you did this in a Future.of? Can you just remove the future wrapping and put the signature calculation in the map call below?

src/frame/worker/tests/GroupCrypto.test.ts
```diff
@@ -100,33 +100,37 @@ describe("GroupCrypto", () => {
...
+            jest.spyOn(Recrypt, "encryptPlaintextToList").mockReturnValue(Future.of(["accessKey1", "accessKey2"]) as any);
+            jest.spyOn(Recrypt, "schnorrSignUtf8String").mockReturnValue(signature as any);
 
             const groupPrivateKey = TestUtils.getTransformedSymmetricKey();
             const adminPrivateKey = new Uint8Array(20);
             const userList = [{id: "user-35", masterPublicKey: {x: "", y: ""}}];
             const signingKeys = TestUtils.getSigningKeyPair();
 
-            GroupCrypto.addAdminsToGroup(groupPrivateKey, userList, adminPrivateKey, signingKeys).engage(
+            GroupCrypto.addAdminsToGroup(groupPrivateKey, groupPublicKey, "grouID", userList, adminPrivateKey, signingKeys).engage(
```
coltfred: nitpick: I know it doesn't matter, but did you mean for `grouID` to be `groupID`?

src/frame/endpoints/GroupApiEndpoints.ts
ernieturner: @coltfred Do you mean writing an integration test that actually tries to hit ironcore-id with an invalid signature? Or are you thinking of something else?

src/frame/worker/tests/GroupCrypto.test.ts
ernieturner: There should probably be an assertion below that this value was used in the expected place. Which probably means a spy on the `Recrypt.schnorrSignUtf8String` method.

src/frame/worker/GroupCrypto.ts
ernieturner: Ya, sorry for the poor guidance on this, I wasn't thinking it through. Do what Colt said.

src/frame/endpoints/GroupApiEndpoints.ts
coltfred: No, I thought you had tests that validated the structure that was passed to `fetch`. That test ends up looking pretty dumb as you just assert that the same structure you just produced is sent.

The best test I have for this in ironweb is to send an invalid signature and verify that it fails from the test app and then fix it and verify it succeeds. We did something similar in ironoxide.

ernieturner: Right, [my comment below](https://github.com/IronCoreLabs/ironweb/pull/58/files#r343837747) addressed that problem here that we weren't asserting that the signature field was part of the call to fetch.

ToniaSchlick: Fixed these issues. I also tested locally that if I call addAdmins with an invalid signature the action will fail.

src/frame/endpoints/GroupApiEndpoints.ts
ToniaSchlick: fixed it

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
ToniaSchlick: done

src/frame/worker/GroupCrypto.ts

src/frame/worker/tests/GroupCrypto.test.ts
ToniaSchlick: done

PR #59, opened 2019-11-08
added needsRotation to group create request

ToniaSchlick:


src/shim/sdk/GroupSDK.ts
```diff
@@ -46,6 +46,7 @@ export function create(options: GroupCreateOptions = {groupName: "", addAsMember
             groupID: options.groupID || "",
             groupName: options.groupName || "",
             addAsMember: options.addAsMember !== false,
+            needsRotation: options.needsRotation || false,
```
ernieturner: Minor, but lets change this to `options.needsRotation === true`. That way the value will always be guaranteed to be a boolean no matter if they provide `undefined/null/etc`.

ToniaSchlick: done

PR #61, opened 2019-11-14
Create group with init members

ToniaSchlick:


ToniaSchlick:
I still need to add error handling for the case a user in the user list is not found.

src/frame/sdk/tests/GroupApi.test.ts
```diff
@@ -168,37 +165,86 @@ describe("GroupApi", () => {
...
 
                     expect(GroupApiEndpoints.callGroupCreateApi).toHaveBeenCalledWith("23", "pub", "encGroupKey", "private group", undefined);
-                    expect(GroupOperations.groupCreate).toHaveBeenCalledWith(TestUtils.userPublicBytes, ApiState.signingKeys(), false);
+                    expect(GroupOperations.groupCreate).toHaveBeenCalledWith(TestUtils.userPublicBytes, ApiState.signingKeys(), undefined);
                 }
             );
         });
 
-        it("if addAsMemeber is set to true, transform key will return with a value", () => {
+        it("if addAsMemeber is set to true, mebers list will be created containing the group creator", () => {
```
ToniaSchlick: change meber to "member"

ironweb.d.ts
```diff
@@ -30,6 +30,7 @@ export interface GroupCreateOptions {
     groupID?: string;
     groupName?: string;
     addAsMember?: boolean;
+    userList?: string[];
```
ernieturner: This should be called `memberList` as the current name doesn't make it clear what the users are. Since members and admins are both "users", we need to be specific about what list this is.

src/frame/endpoints/GroupApiEndpoints.ts
```diff
@@ -63,14 +64,13 @@ function groupGet(groupID: string) {
...
-    if (createPayload.transformKey) {
-        memberList = [
-            {
-                userId: createPayload.userID,
-                userMasterPublicKey: userPublicKeyString,
-                transformKey: transformKeyToBase64(createPayload.transformKey),
-            },
-        ];
+
+    if (createPayload.transformKeyGrantList) {
```
ernieturner: When you make the change to have `transformKeyGrantList` not optional we should make this more clear what we're checking here, change this to 
```suggestion
    if (createPayload.transformKeyGrantList.length > 0) {
```

src/WorkerMessageTypes.d.ts
```diff
@@ -173,16 +172,16 @@ export interface GroupCreateWorkerRequest {
...
+        memberList?: UserOrGroupPublicKey[];
     };
 }
 export interface GroupCreateWorkerResponse {
     type: "GROUP_CREATE_RESPONSE";
     message: {
         encryptedGroupKey: PREEncryptedMessage;
         groupPublicKey: PublicKey<Uint8Array>;
-        transformKey?: TransformKey;
+        transformKeyGrantList?: TransformKeyGrant[];
```
ernieturner: Lets change all of the places that take `transformKeyGrantList` as optional to just be required. Because it can be provided as an empty array and still be valid for the type, lets make it concrete. This change should spider out into various places including the request layer.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
...
  * @param {boolean} addAsMember Whether the group creator shoudl be added as a member upon create
  */
-export function create(groupID: string, groupName: string, addAsMember: boolean) {
-    return GroupOperations.groupCreate(ApiState.userPublicKey(), ApiState.signingKeys(), addAsMember)
-        .flatMap(({encryptedGroupKey, groupPublicKey, transformKey}) =>
-            GroupApiEndpoints.callGroupCreateApi(groupID, groupPublicKey, encryptedGroupKey, groupName, transformKey)
-        )
-        .map((createdGroup) => formatDetailedGroupResponse(createdGroup) as GroupDetailResponse);
+export function create(groupID: string, groupName: string, addAsMember: boolean, userList?: string[]): Future<SDKError, GroupDetailResponse> {
+    const creator = {id: ApiState.user().id, masterPublicKey: publicKeyToBase64(ApiState.userPublicKey())};
```
ernieturner: I'd just move this `creator` object into the line below and only create it if you need to.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
     return GroupApiEndpoints.callGroupGetApi(groupID).map(formatDetailedGroupResponse);
 }
 
+function createErrorForInvalidUserList(userKeys: UserKeyListResponseType, userGrants: string[]) {
+    const existingUserIDs = userKeys.result.map(({id}) => id);
+    const missingUsersString = userGrants.filter((userID) => existingUserIDs.indexOf(userID) === -1);
+    return missingUsersString;
```
ernieturner: Remove this and just do `return userGrants.filter(....`

ernieturner: Also, isn't `missingUsersString` an array not a string?

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
     return GroupApiEndpoints.callGroupGetApi(groupID).map(formatDetailedGroupResponse);
 }
 
+function createErrorForInvalidUserList(userKeys: UserKeyListResponseType, userGrants: string[]) {
+    const existingUserIDs = userKeys.result.map(({id}) => id);
+    const missingUsersString = userGrants.filter((userID) => existingUserIDs.indexOf(userID) === -1);
```
ernieturner: We should use `Array.includes` here as it's more appropriate than `indexOf`.
```suggestion
    const missingUsersString = userGrants.filter((userID) => existingUserIDs.includes(userID));
```
You might see a bunch of other code that uses `indexOf` but that was because `includes` doesn't work in IE. Now that we don't support IE we can use `includes` instead.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
...
+                      return Future.reject(
+                          new SDKError(
+                              new Error(`Failed to create group due to unknown users in user list. Missing user IDs: [${missingUsersString}]`),
+                              ErrorCodes.GROUP_CREATE_WITH_MEMBERS_FAILURE
+                          )
+                      );
+                  }
+                  return Future.of(userKeysResponse.result.map((user) => ({id: user.id, masterPublicKey: user.userMasterPublicKey})));
+              })
+            : Future.of([]);
```
ernieturner: I'd move this `userList` calculation into it's own private function in this file.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
     return GroupApiEndpoints.callGroupGetApi(groupID).map(formatDetailedGroupResponse);
 }
 
+function createErrorForInvalidUserList(userKeys: UserKeyListResponseType, userGrants: string[]) {
```
ernieturner: Add a comment to this function to explain what it's doing.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
...
  */
-export function create(groupID: string, groupName: string, addAsMember: boolean) {
-    return GroupOperations.groupCreate(ApiState.userPublicKey(), ApiState.signingKeys(), addAsMember)
-        .flatMap(({encryptedGroupKey, groupPublicKey, transformKey}) =>
-            GroupApiEndpoints.callGroupCreateApi(groupID, groupPublicKey, encryptedGroupKey, groupName, transformKey)
-        )
-        .map((createdGroup) => formatDetailedGroupResponse(createdGroup) as GroupDetailResponse);
+export function create(groupID: string, groupName: string, addAsMember: boolean, userList?: string[]): Future<SDKError, GroupDetailResponse> {
+    const creator = {id: ApiState.user().id, masterPublicKey: publicKeyToBase64(ApiState.userPublicKey())};
+    const creatorMemberKeys = addAsMember ? [creator] : [];
```
ernieturner:
```suggestion
    const creatorMemberKey = addAsMember ? [creator] : [];
```
Since it's only a single key

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
...
+                              ErrorCodes.GROUP_CREATE_WITH_MEMBERS_FAILURE
+                          )
+                      );
+                  }
+                  return Future.of(userKeysResponse.result.map((user) => ({id: user.id, masterPublicKey: user.userMasterPublicKey})));
+              })
+            : Future.of([]);
+
+    return userListResult
+        .map((userMemberKeys) => [...creatorMemberKeys, ...userMemberKeys])
```
ernieturner: I'd remove this `.map` and change the logic in the flatMap below. Since `memberKeys` below will always be an array, you can then just do `memberKeys.concat(creatorMemberKey)` as the 3rd parameter to `groupCreate`. Similar to what I mentioned above, once we have that argument always be an array and not an optional array, this change will fit in nicely there.

src/frame/sdk/GroupApi.ts
```diff
@@ -84,18 +85,46 @@ export function get(groupID: string) {
...
+        .flatMap((memberKeys) =>
+            GroupOperations.groupCreate(ApiState.userPublicKey(), ApiState.signingKeys(), addAsMember || userList !== undefined ? memberKeys : undefined)
+                .flatMap(({encryptedGroupKey, groupPublicKey, transformKeyGrantList}) =>
```
ernieturner: Can't this `flatMap` call move up a level and not be nested under the `memberKeys` flatMap? Pretty sure this operation can all just be a single level of `flatMap`s.

src/frame/sdk/GroupOperations.ts
```diff
@@ -7,10 +7,14 @@ import * as WMT from "../../WorkerMessageTypes";
  * @param {SigningKeyPair}        signingKeys   Current users signing keys
  * @param {boolean}               addAsMember   Whether to also add user who is creating group as a member of the group
  */
-export function groupCreate(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, addAsMember: boolean) {
+export function groupCreate(
+    userPublicKey: PublicKey<Uint8Array>,
+    signingKeys: SigningKeyPair,
+    memberList?: {id: string; masterPublicKey: PublicKey<string>}[]
```
ernieturner: Similar here, this should just be a fixed type and be an empty array if there are no member keys.

src/frame/sdk/GroupApi.ts
ToniaSchlick: missed this one I will get it.

src/frame/sdk/GroupApi.ts
```diff
@@ -86,17 +87,56 @@ export function get(groupID: string) {
     return GroupApiEndpoints.callGroupGetApi(groupID).map(formatDetailedGroupResponse);
 }
 
+/**
+ * Create the error strings for missing users
+ */
+function createErrorForInvalidUserList(userKeysResponse: UserKeyListResponseType, memberList: string[]) {
```
ernieturner:
```suggestion
function getListOfMissingUsersFromKeyResponse(userKeysResponse: UserKeyListResponseType, memberList: string[]) {
```

src/frame/sdk/GroupApi.ts
```diff
@@ -86,17 +87,56 @@ export function get(groupID: string) {
...
+    const existingUserIDs = userKeysResponse.result.map(({id}) => id);
+    return memberList.filter((userID) => !existingUserIDs.includes(userID));
+}
+
+/**
+ * Checkes that all members listed in member list are existing members
+ */
+function validateMemberList(memberList: string[]) {
```
ernieturner: I expect that when we do the add admins on create work this method will be used for both, so a more generic name will be better going forward.
```suggestion
function getCompleteListOfUserPublicKeys(memberList: string[]) {
```

src/frame/sdk/GroupApi.ts
```diff
@@ -86,17 +87,56 @@ export function get(groupID: string) {
...
+    return memberList !== undefined && memberList.length > 0
```
ernieturner: If the `memberList` parameter can be undefined it should be marked as such in the parameter types. So `(memberList?: string[])`.

src/frame/sdk/GroupApi.ts
ToniaSchlick: done

src/frame/sdk/GroupApi.ts
```diff
@@ -86,17 +87,56 @@ export function get(groupID: string) {
...
+        ? UserApiEndpoints.callUserKeyListApi(memberList).flatMap((userKeysResponse) => {
+              if (userKeysResponse.result.length !== memberList.length) {
+                  const missingUsersString = createErrorForInvalidUserList(userKeysResponse, memberList);
+                  return Future.reject(
+                      new SDKError(
+                          new Error(`Failed to create group due to unknown users in user list. Missing user IDs: [${missingUsersString}]`),
```
ernieturner: To confirm, isn't `missingUsersString` an array at this point? If so, we should rename it and then call `missingUsers.join(", ")` to convert it to a string instead of relying on whatever JavaScript decides to do to `toString` the array here.

src/frame/sdk/GroupApi.ts
```diff
@@ -86,17 +87,55 @@ export function get(groupID: string) {
...
-            GroupApiEndpoints.callGroupCreateApi(groupID, groupPublicKey, encryptedGroupKey, needsRotation, groupName, transformKey)
+export function create(
+    groupID: string,
+    groupName: string,
+    addAsMember: boolean,
+    needsRotation: boolean,
+    memberList?: string[]
+): Future<SDKError, GroupDetailResponse> {
+    const creatorMemberKey = addAsMember ? [{id: ApiState.user().id, masterPublicKey: publicKeyToBase64(ApiState.userPublicKey())}] : [];
+    const memberListValidationResult = memberList !== undefined && memberList.length > 0 ? validateMemberList(memberList) : Future.of([]);
```
ernieturner: The `validateMemberList` method is already checking if the provided list is undefined, so you could probably just leave that here and then instead of creating the `memberListValidationResult` variable, just immediately do 

```
return validateMemberList(memberList)
	.flatMap((memberKeys) => ...
```

src/frame/sdk/GroupOperations.ts
```diff
@@ -3,14 +3,14 @@ import * as WMT from "../../WorkerMessageTypes";
...
  * Send message to worker to create a new group.
- * @param {PublicKey<Uint8Array>} userPublicKey Users public key to encrypt document key to. The user who is creating this group will become the first group admin
- * @param {SigningKeyPair}        signingKeys   Current users signing keys
- * @param {boolean}               addAsMember   Whether to also add user who is creating group as a member of the group
+ * @param {PublicKey<Uint8Array>}       userPublicKey Users public key to encrypt document key to. The user who is creating this group will become the first group admin
+ * @param {SigningKeyPair}              signingKeys   Current users signing keys
+ * @param {{id, PublicKey<string>}[]}   memberList    List of users to be initalized as members when the group is created
  */
-export function groupCreate(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, addAsMember: boolean) {
+export function groupCreate(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, memberList: {id: string; masterPublicKey: PublicKey<string>}[]) {
```
ernieturner: You can use the type `UserOrGroupPublicKey` here for the `memberList` type.

src/frame/worker/GroupCrypto.ts
```diff
@@ -11,17 +11,17 @@ import {publicKeyToBytes} from "../../lib/Utils";
  * @param {SigningKeyPair}        signingKeys   Current users signing keys used to sign transform key
  * @param {boolean}               addAsMember   Whether to add the current group admin as a member
  */
-export function createGroup(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, addAsMember: boolean) {
+export function createGroup(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, memberList?: UserOrGroupPublicKey[]) {
```
ernieturner: Similar to the `transformKeyGrantList` change to make that a fixed array, it's probably beneficial to do the same thing here and make this `memberList` not be optional. That'll clean up the code below to not have to check whether it's undefined or not. That change should also spider out to the message passing that happens between the frame and the worker but I don't think it'll change the code that builds up that message because it's already always being passed as an array, it just might be empty.

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
```diff
@@ -184,16 +191,20 @@ describe("GroupApiEndpoints", () => {
...
+                y: new Uint8Array([110, 98]),
+            };
             const groupEncryptedPrivateKey: any = {
                 encryptedMessage: "abc",
                 authorizationCode: "auth",
                 ephemeralPublicKey: "epub",
             };
 
-            const transformKey = undefined;
+            // const transformKeyGrantList = [];
```
ernieturner: Remove?

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
```diff
@@ -184,16 +191,20 @@ describe("GroupApiEndpoints", () => {
...
 
-            GroupApiEndpoints.callGroupCreateApi("", groupPublicKey, groupEncryptedPrivateKey, false, "group name", transformKey).engage(
+            GroupApiEndpoints.callGroupCreateApi("", groupPublicKey, groupEncryptedPrivateKey, false, [], "group name").engage(
```
ernieturner: It'd probably be good to add a test where we pass in two transform keys to make sure that the request that is sent includes them both.

src/frame/tests/index.test.ts
```diff
@@ -632,15 +632,15 @@ describe("frame index", () => {
 
             const payload: MT.GroupCreateRequest = {
                 type: "GROUP_CREATE",
-                message: {groupID: "my-group", groupName: "bar", addAsMember: true, needsRotation: false},
+                message: {groupID: "my-group", groupName: "bar", addAsMember: true, needsRotation: false, memberList: ["string"]},
```
ernieturner: Very minor, but I'd probably change "string" to something like "userID" so that it's a lot more clear about what type of data you're passing.

src/frame/worker/GroupCrypto.ts
```diff
@@ -11,17 +11,17 @@ import {publicKeyToBytes} from "../../lib/Utils";
...
-export function createGroup(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, addAsMember: boolean) {
+export function createGroup(userPublicKey: PublicKey<Uint8Array>, signingKeys: SigningKeyPair, memberList: UserOrGroupPublicKey[]) {
     return loadRecrypt()
         .flatMap((Recrypt) => {
             return Recrypt.generateGroupKeyPair().flatMap(({publicKey, plaintext, privateKey}) => {
                 return Future.gather2(
                     Recrypt.encryptPlaintext(plaintext, userPublicKey, signingKeys),
-                    addAsMember ? Recrypt.generateTransformKey(privateKey, userPublicKey, signingKeys) : Future.of(undefined)
-                ).map(([encryptedGroupKey, transformKey]) => ({
+                    memberList.length > 0 ? Recrypt.generateTransformKeyToList(privateKey, memberList, signingKeys) : Future.of([])
```
ernieturner: You can just call `generateTransformKeyToList` with the `memberList` directly because it already checks if the provided array has a length before doing anything.

src/frame/tests/index.test.ts
ToniaSchlick: done

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
ToniaSchlick: I added this check to the test "combines all group content into payload and maps response to data result".

src/frame/worker/GroupCrypto.ts

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
```diff
@@ -81,16 +81,26 @@ describe("GroupApiEndpoints", () => {
...
         it("combines all group content into payload and maps response to data result", () => {
+            const groupPublicKey = {
+                x: new Uint8Array([98, 105, 133]),
+                y: new Uint8Array([110, 98]),
+            };
+
             const groupEncryptedPrivateKey = TestUtils.getEncryptedSymmetricKey();
-            const transformKey = TestUtils.getTransformKey();
+            const creator = TestUtils.getFullUser();
+            // const transformKeyGrant = {id: creator.id, publicKey: creator.userMasterPublicKey, transformKey: TestUtils.getTransformKey()};
```
ernieturner: Remove

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
```diff
@@ -81,16 +81,26 @@ describe("GroupApiEndpoints", () => {
...
 
-            GroupApiEndpoints.callGroupCreateApi("35", groupPublicKey, groupEncryptedPrivateKey, false, "group name", transformKey).engage(
+            GroupApiEndpoints.callGroupCreateApi("35", groupPublicKey, groupEncryptedPrivateKey, false, transformKeyGrantList, "group name").engage(
                 (e) => fail(e),
                 (group: any) => {
                     expect(group).toEqual({foo: "bar"});
                     expect(ApiRequest.makeAuthorizedApiRequest).toHaveBeenCalledWith("groups", expect.any(Number), expect.any(Object));
                     const request = (ApiRequest.makeAuthorizedApiRequest as jasmine.Spy).calls.argsFor(0)[2];
-
+                    console.log(request.body);
```
ernieturner: Remove

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
ToniaSchlick: done

src/frame/endpoints/tests/GroupApiEndpoints.test.ts
ToniaSchlick: done

PR #65, opened 2019-11-27
Added support for group private key rotation

ToniaSchlick:


src/frame/sdk/GroupOperations.ts
```diff
@@ -15,6 +15,27 @@ export function groupCreate(signingKeys: SigningKeyPair, memberList: UserOrGroup
     return WorkerMediator.sendMessage<WMT.GroupCreateWorkerResponse>(payload).map(({message}) => message);
 }
 
+/**
+ * Send message to the worker to rotate the group private key and encrypt the new private key to the provided list of admins
+ */
+export function rotateAndEncryptNewGroupPrivateKeyToList(
+    encryptedGroupKey: TransformedEncryptedMessage,
+    adminList: UserOrGroupPublicKey[],
+    myPrivateKey: PrivateKey<Uint8Array>,
```
ernieturner: Maybe just `userPrivateMasterKey`

src/frame/worker/GroupCrypto.ts
```diff
@@ -27,6 +27,36 @@ export function createGroup(signingKeys: SigningKeyPair, memberList: UserOrGroup
         })
         .errorMap((error) => new SDKError(error, ErrorCodes.GROUP_KEY_GENERATION_FAILURE));
 }
+/**
+ *  Rotate current group private key. Then encrypts rotated private to the groups admins.
+ */
+export function rotatePrivateKey(
+    encryptedGroupKey: TransformedEncryptedMessage,
+    adminList: UserOrGroupPublicKey[],
+    adminPrivateKey: PrivateKey<Uint8Array>,
```
ernieturner: Ditto here from above with the name of this key

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -125,13 +125,39 @@ const rotateUsersPrivateKey = (userPrivateKey: Uint8Array): Future<Error, {newPr
...
-
 export const rotateUsersPrivateKeyWithRetry = (userPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
     return rotateUsersPrivateKey(userPrivateKey).handleWith(() => rotateUsersPrivateKey(userPrivateKey));
 };
 
+/**
+ * Generates a new key pair and takes the private key as the new group private key, The existing private key is then subtracted from the newly
+ * generated private key as the augmentation factor.
+ */
+const rotateGroupPrivateKey = (ExistingGroupPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
```
ernieturner:
```suggestion
const rotateGroupPrivateKey = (existingGroupPrivateKey: Uint8Array): Future<Error, {newPrivateKey: Uint8Array; augmentationFactor: Uint8Array}> => {
```

src/frame/worker/crypto/recrypt/RecryptWasm.ts
```diff
@@ -125,13 +125,39 @@ const rotateUsersPrivateKey = (userPrivateKey: Uint8Array): Future<Error, {newPr
...
+    return generateKeyPair().flatMap(({privateKey}) => {
+        if (isBufferAllZero(privateKey)) {
+            return Future.reject(new Error("Key rotation failed."));
+        }
+        const newPrivateKey = RecryptApi.hash256(privateKey);
```
ernieturner: Why is this necessary here to rotate a groups private key but it isn't necessary when we do user key rotation?

integration/components/groups/GroupDetail.tsx
```diff
@@ -133,6 +137,40 @@ export default class GroupDetail extends React.Component<GroupDetailProps, Group
...
+    /**
+     * Verify the group ID the user typed is correct before making the actual delete call.
+     */
+    roatateGroupPrivateKey = () => {
+        IronWeb.group
+            .rotatePrivateKey(this.props.group.groupID)
+            .then((groupKeyUpdate) => {
+                logAction(`Group private key rotation was successful`);
+                this.setState({needsRotation: groupKeyUpdate.needsRotation});
+                this.getRotateGroupPrivateKeyIcon();
```
ernieturner: There's no need to call this method here. It's not doing anything. With React, the `render()` function automatically gets called after you call `this.setState` so that the UI reflects the new state.

integration/components/groups/GroupDetail.tsx
```diff
@@ -133,6 +137,40 @@ export default class GroupDetail extends React.Component<GroupDetailProps, Group
...
+                    <FloatingActionButton onClick={this.roatateGroupPrivateKey} mini backgroundColor={red700}>
+                        <ImageRotateRight />
+                    </FloatingActionButton>
+                );
+            } else
+                return (
+                    <FloatingActionButton onClick={this.roatateGroupPrivateKey} mini backgroundColor={grey400}>
+                        <ImageRotateRight />
+                    </FloatingActionButton>
+                );
```
ernieturner: The formatting here is weird because you're using brackets for the if, but not for the else. Instead I'd just remove the else and do

```
if(){
   return..
}
return ..
```

integration/components/groups/NewGroup.tsx
```diff
@@ -24,7 +25,8 @@ export default class NewGroup extends React.Component<NewGroupProps, NewGroupSta
     constructor(props: NewGroupProps) {
         super(props);
         this.state = {
-            isChecked: true,
+            isAddAsMemberChecked: true,
+            isNeedsRotationChecked: true,
```
ernieturner: This should default to false because that's the default in the SDK.

src/frame/sdk/GroupApi.ts
```diff
@@ -145,6 +152,33 @@ export function create(
...
+            );
+        }
+        return UserApiEndpoints.callUserKeyListApi(group.adminIds)
+            .map((adminKeys) => adminKeys.result.map((user) => ({id: user.id, masterPublicKey: user.userMasterPublicKey})))
+            .flatMap((adminKeys) =>
+                GroupOperations.rotateGroupPrivateKeyAndEncryptToAdmins(group.encryptedPrivateKey, adminKeys, privateKey, ApiState.signingKeys())
+            )
+            .flatMap(({encryptedAccessKeys, augmentationFactor}) =>
+                GroupApiEndpoints.callGroupPrivateKeyUpdateApi(groupID, encryptedAccessKeys, augmentationFactor, group.currentKeyId).map((result) => ({
+                    needsRotation: result.needsRotation,
```
ernieturner: You can simplify this a little bit by destructing and doing 
```
.map(({needsRotation}) => ({needsRotation}))
```

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -52,6 +52,48 @@ describe("RecryptWasm", () => {
         });
     });
 
+    describe("rotateGroupPrivateKeyWithRetry", () => {
+        const groupPrivateKey = new Uint8Array([22, 33, 44]);
+        it("should result in an error when mocked generateKeyPair returns Uint8Array of zeros for new privateKey", () => {
+            jest.spyOn(Recrypt.getApi(), "generatePlaintext");
```
ernieturner: Should this spy be returning something? I looks like this test is meant to verify that if the plaintext we generate is all zeros it should fail, but it looks like here we're not generating anything? It should probably have a `.mockReturnValue(new Uint8Array(32))`.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -52,6 +52,48 @@ describe("RecryptWasm", () => {
...
+            Recrypt.rotateGroupPrivateKeyWithRetry(groupPrivateKey).engage(
+                (error) => {
+                    expect(error.message).toEqual("Key rotation failed.");
```
ernieturner: I'd just remove this line. There's little to no benefit in having to update this test if we decide to change the text of the error message. The assert below this is sufficient.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -52,6 +52,48 @@ describe("RecryptWasm", () => {
...
+                () => fail("Should not success when operation fails")
+            );
+        });
+
+        it("should result in an error when subtracting existing private key from the new private key", () => {
+            jest.spyOn(Recrypt.getApi(), "generatePlaintext").mockReturnValue(new Uint8Array([12, 23, 34]) as any);
+            jest.spyOn(MockRecrypt, "subtractPrivateKeys").mockReturnValue(new Uint8Array(32));
+            Recrypt.rotateGroupPrivateKeyWithRetry(groupPrivateKey).engage(
+                (error) => {
+                    expect(error.message).toEqual("Key rotation failed.");
```
ernieturner: Similar to above, it'd be better to have some assertion here about the calls to `generatePlaintext` or `subtractPrivateKeys` instead of asserting this exact error message text.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
```diff
@@ -52,6 +52,48 @@ describe("RecryptWasm", () => {
...
+                },
+                () => fail("Should fail when private key subtraction results in zeroed private key")
+            );
+
+            expect(MockRecrypt.subtractPrivateKeys).toHaveBeenCalledTimes(2);
```
ernieturner: This needs to be moved up into the error callback.

src/frame/sdk/GroupApi.ts
```diff
@@ -145,6 +152,33 @@ export function create(
...
+        if (!isGroupAdminResponse(group)) {
+            return Future.reject(
+                new SDKError(
+                    new Error("Current user is not authorized to rotate this group's private key as they are not a group administrator."),
+                    ErrorCodes.GROUP_PRIVATE_KEY_ROTATION_FAILURE
+                )
+            );
+        }
+        return UserApiEndpoints.callUserKeyListApi(group.adminIds)
+            .map((adminKeys) => adminKeys.result.map((user) => ({id: user.id, masterPublicKey: user.userMasterPublicKey})))
```
ernieturner: We should probably add a check here to make sure that we got back the same number of public keys that we requested. If there's 10 admins of this group and we only get 7 public keys back from this call, that should probably cause a failure of the rotation because the group membership won't be the same before/after key rotation.

coltfred: @ernieturner The server will reject if the admins list doesn't match what's in the database. You can still check if you want, but shouldn't ever happen.

ernieturner: Oh neat, good to know. Then ya, I'm fine with not having a check for that.

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: In the rycrypt-wasm-binding  generatePlaintext is mocked for jest or other unit to return new Uint8Array(384). Do I still need to mock it here?

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ToniaSchlick: done

src/frame/worker/crypto/recrypt/tests/RecryptWasm.test.ts
ernieturner: Oh right I forgot about the mock here. But when you re-mock it like you're doing here, it will override the other one. So I'm pretty sure it's not returning anything here. We should make it be explicit and have it return what you want based on what you're testing.

ToniaSchlick: ok, done

integration/components/groups/GroupDetail.tsx
```diff
@@ -133,6 +137,39 @@ export default class GroupDetail extends React.Component<GroupDetailProps, Group
...
+                return (
+                    <FloatingActionButton onClick={this.roatateGroupPrivateKey} mini backgroundColor={red700}>
+                        <ImageRotateRight />
+                    </FloatingActionButton>
+                );
+            }
+            return (
+                <FloatingActionButton onClick={this.roatateGroupPrivateKey} mini backgroundColor={grey400}>
+                    <ImageRotateRight />
+                </FloatingActionButton>
```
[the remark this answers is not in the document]
ToniaSchlick: done

PR #69, opened 2019-12-17
Create new device response

ToniaSchlick:


ironweb.d.ts
```diff
@@ -139,11 +139,14 @@ export interface UserCreateResponse {
     userMasterPublicKey: PublicKey<Base64String>;
     needsRotation: boolean;
 }
-export interface DeviceKeys {
+export interface DeviceAddResponse {
```
ernieturner: We can't change the name of this interface without doing a major version bump because people could be importing this type into their own code and it would break on upgrade. I don't feel that this rename is enough of a reason to break things, so I'd just revert this.

ToniaSchlick: ok, done

Issue #74, opened 2020-03-16
Expose substring search functionality


BobWall23:
- [ ] Consume the new recrypt-wasm-binding that includes the functions to tokenize a string for indexing and for querying
- [ ] Implement the ability to create a search index, initialize a search index using an encrypted value, tokenize data and tokenize query (see https://github.com/IronCoreLabs/ironoxide/blob/master/src/search.rs)

These functions should exist on a namespace on the SDK object.

The search indices will be identified by handles - creating the index generates and encrypts a salt and returns that along with a handle. The initialization function takes an encrypted salt and returns the handle. Inside the frame, there should be a map from handle to decrypted salt. The tokenize calls should also take the handle.

PR #75, opened 2020-03-30
#74 Add methods to allow users to create and use a encrypted blind se…

ernieturner:
…arch index

- [x] Unit tests
- [x] Version bump (minor)
- [x] Changelog entry


CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The IronWeb SDK NPM releases follow standard [Semantic Versioning](https://semve
...
+## v3.4.1
+
+Encrypted search index feature. Adds methods to create a blind encrypted search index that allows for searching over encrypted data.
+
++ New namespace, `IronWeb.search`, that adds methods to create and initialize a blind search index and then methods for tokenizing data and search queries.
+  + A new blind search index can be created via the `IronWeb.search.createBlindSearchIndex(groupId: string): BlindIndex` to create a new search index that is accessible by everyone in the provided group.
+  + Initialize an existing search index via `IronWeb.search.initializeBlindSearchIndex(index: BlindIndex): InitializedSearchIndex` which takes the result of the `createBlindSearchIndex` call and returns an `InitializedSearchIndex` which contains methods for tokenizing data
+  + The `tokenizeData` method on `InitializedSearchIndex` creates an index for the provided plaintext data. The resulting `Uint32Array` is the index of all the trigrams in the provided plaintext.
+  + The `tokenizeQuery` method on `InitializedSearchIndex` tokenizes the provided search query into trigrams and returns a `Uint32Array` which can be compared against the result of `tokenizeData` to find matches.
+
```
ernieturner: @BobWall23 @coltfred I could use some review on this changelog as I feel like I'm sort of winging it on my descriptions. I don't want to add too much content here and we should add the "Here's more info" and link to the content on the website that we'll eventually have.

src/frame/sdk/SearchApi.ts
```diff
@@ -0,0 +1,90 @@
...
+            salt: searchIndexCache[searchIndexId],
+        },
+    };
+    return WorkerMediator.sendMessage<WMT.SearchTokenizeStringResponse>(payload).map(({message}) => message);
+};
+
+/**
+ * Pass provided string to WebWorker for transliteration
+ */
+export const transliterateString = (string: string) => {
```
clintfred, line 84: Why/when is this needed?

src/frame/FrameUtils.ts
```diff
@@ -144,3 +144,8 @@ export function combineDocumentParts(documentID: string, segmentID: number, docu
 export function encryptedDocumentToBase64(documentID: string, segmentID: number, document: EncryptedDocument) {
     return fromByteArray(combineDocumentParts(documentID, segmentID, document));
 }
+
+/**
+ * Encode the provided bytes as hex
+ */
+export const encodeBytesAsHex = (byteSource: Uint8Array): string => Array.prototype.map.call(byteSource, (byte) => `00${byte.toString(16)}`.slice(-2)).join("");
```
skeet70, line 151: I googled this trying to understand it and the big [SO hit](https://stackoverflow.com/questions/34309988/byte-array-to-hex-string-conversion-in-javascript) seems to have problems with this approach? I don't really understand it well enough to know if it's a problem, but thought I'd leave it just in case.

src/frame/sdk/SearchApi.ts
ernieturner: It's hopefully described in the ticket that [added this method](https://github.com/IronCoreLabs/search-helpers/issues/9) to `search-helpers`, but if that doesn't describe it enough, let me know.

src/frame/FrameUtils.ts
ernieturner: I think this code is actually OK because of the `Array.prototype.map.call` and not just doing `Uint8Array.map`, but the `Array.from` syntax looks a lot nicer, so I just ended up implementing it that way.

PR #86, opened 2020-12-14
GitHub actions

skeet70:
I'm using the `GITHUB_TOKEN` in the action, which should allow for writing to the repo. Added an `NPM_TOKEN` secret which should be able to do the needed publishing.


coltfred (approved):
:crossed_fingers:

cjyar (commented):
Looks good. Any reason not to delete `.travis.yml`?

PR #97, opened 2021-06-01
Remove MS Legacy Edge support

clintfred:
And update version of recrypt-wasm-binding

There was also some cruft left over from when IE 11 support was removed a while back. That is also removed here.

Tested on the following via browserstack

| Platform | Browser | Notes | Worked? |
| ----------- | ----------- | ------- | ------------ |
| Mac/Big Sur | Chrome 85 | | YES |
| Mac/High Sierra | Safari 11.1 | | YES |
| Win 10 | Edge 18 | Error code 206 on submit password | NO |
| Win 10 | Edge 91 |  | YES |
| Win 10 | IE 11 | Blank page. Console: "Promise is undefined" | NO |
| Win 10 | Safari 5.1 | Blank page | NO |
| Win 10 | Opera 76 | | YES |
| Win 10 | Yandex.12 | UI comes up. Initialize fails with "No WASM support" | NO |
| Win 7 | IE 11 | Blank page. Console: "Promise is undefined" | NO |
| iOS/iPhone 12 | Safari | | YES |
| Android/S8 | Android Browser | | YES |

### DIFF
--- a/src/frame/worker/crypto/aes/tests/index.test.ts
+++ b/src/frame/worker/crypto/aes/tests/index.test.ts
@@ -131,22 +131,6 @@ describe("AES", () => {
             );
         });
 
-        it("uses polyfill API if native is not available", () => {
-            spyOn(PolyfillAes, "decryptDocument").and.returnValue(Future.of({foo: "bar"}));
-
-            const doc = new Uint8Array(40);
-            const symKey = new Uint8Array(32);
-            const nonce = new Uint8Array(12);
-
-            decryptDocument(doc, symKey, nonce).engage(
-                () => fail("document decryption should not fail"),
-                (decryptedDoc: any) => {
-                    expect(decryptedDoc).toEqual({foo: "bar"});
-                    expect(PolyfillAes.decryptDocument).toHaveBeenCalledWith(doc, symKey, nonce);
-                }
-            );
-        });
-
         it("falls back to polyfill API if native fails", () => {
             spyOn(PolyfillAes, "decryptDocument").and.returnValue(Future.of({foo: "bar"}));
             spyOn(NativeAes, "decryptDocument").and.returnValue(Future.reject(new Error("forced failure")));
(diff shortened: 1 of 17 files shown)


src/frame/worker/crypto/aes/tests/index.test.ts
```diff
@@ -276,7 +264,6 @@ describe("AES", () => {
         });
 
         it("uses polyfill API if native is not available", () => {
-            spyOn(CryptoUtils, "isNativeCryptoSupported").and.returnValue(false);
```
coltfred, line 279: Seems like this test should be defunct right?

src/frame/worker/crypto/aes/tests/index.test.ts
```diff
@@ -212,7 +203,6 @@ describe("AES", () => {
...
```
coltfred, line 215: Seems like this test should be defunct right?

src/frame/worker/crypto/aes/tests/index.test.ts
```diff
@@ -137,7 +132,6 @@ describe("AES", () => {
...
```
coltfred, line 140: Seems like this test should be defunct right?

coltfred (changes requested):
It seems like a few tests that were using the `isNativeCryptoSupported` as false should be removed as well as the commented out code that @giarc3 pointed out.

clintfred:
@cjyar Is there an easy way to remove the build caches? I made a new version of the (temporary) recrypt-wasm-binding-0.6.0 zip file, and it not happy. If not I can upload a new one with a different name.

clintfred:
> @cjyar Is there an easy way to remove the build caches? I made a new version of the (temporary) recrypt-wasm-binding-0.6.0 zip file, and it not happy. If not I can upload a new one with a different name.

I uploaded a new artifact, so that should allow it to build now.

cjyar:
> @cjyar Is there an easy way to remove the build caches? I made a new version of the (temporary) recrypt-wasm-binding-0.6.0 zip file, and it not happy. If not I can upload a new one with a different name.

I don't know what build cache it would be using. I don't see any caching actions in the workflows. Are you talking about this error?

> error https://drive.google.com/uc?export=download&id=1gSv3s59BCo38wASfp3_-P2vhPiz0psLY: Integrity check failed for "@ironcorelabs/recrypt-wasm-binding" (computed integrity doesn't match our records, got "sha512-D0Nb4MJdhEJRFbu3j1cU8p5b4lissWiQ2qvzHM8SzuvDpu3V95FmZhjoKS0faamPqY9d58JzCY4A62DvBx1tCQ== sha1-9ciguH3Z6WevJP5gxpSwW0LwYDA=")

That looks like a yarn error that could be caused by a few different things. I don't think our GitHub runners have a local cache, so it's more likely a problem with `yarn.lock`. [This post](https://www.seancdavis.com/blog/fix-yarn-integrity-check-failed/) has some advice.

skeet70 (approved):
No notes, good ripping.

PR #99, opened 2021-06-04
Remove symbols from build.js that are gone now.

BobWall23:
Murph deleted a bunch of code and updated to webpack 5, which supports default async wasm module loading. Webpack 5 had standardized a bunch of the things we were doing, so webpack configs shrunk and shifted a bit. For ex there was a big comment about how we were doing something to prevent node crypto from being polyfilled, and they just stopped doing any default polyfills.

End result is that none of the "hacky hack" in the `build.js` is necessary anymore and the import for the recrypt wrapper becomes a normal one.


clintfred (changes requested):
This still failed locally for me:

```
$ ./build.js 
src/frame/worker/crypto/recrypt/index.ts:3:13 - error TS2300: Duplicate identifier 'Recrypt'.

3 import * as Recrypt from "./RecryptWasm";
              ~~~~~~~

src/frame/worker/crypto/recrypt/index.ts:4:13 - error TS2300: Duplicate identifier 'Recrypt'.

4 import * as Recrypt from "./RecryptWasm";
              ~~~~~~~

src/frame/worker/crypto/recrypt/index.ts:22:1 - error TS2304: Cannot find name 'getCryptoSubtleApi'.

22 getCryptoSubtleApi; randomSeed; export default function loadRecrypt(): Future<Error, typeof Recrypt> {return Future.of(Recrypt);}
   ~~~~~~~~~~~~~~~~~~

src/frame/worker/crypto/recrypt/index.ts:22:21 - error TS2304: Cannot find name 'randomSeed'.

22 getCryptoSubtleApi; randomSeed; export default function loadRecrypt(): Future<Error, typeof Recrypt> {return Future.of(Recrypt);}
                       ~~~~~~~~~~


Found 4 errors.

/workspace/ironweb/node_modules/shelljs/src/common.js:401
      if (config.fatal) throw e;
```

BobWall23:
@clintfred I believe that if you do `git restore src/frame/worker/crypto/recrypt/index.ts` to restore the file that was left changed by the failed run of `build.js`, it will work. That fixed it for me, anyway

skeet70:
I ran tests and ran the integration app to test, and haven't run into any problems. I didn't run whatever browserstack tests there are out there.

clintfred:
Here are the results of my browserstack testing. Things mostly worked the same. I'm not sure why I couldn't upload an image using Safari 11.1 (I also tried 13.1). Probably something with the demo app and not ironweb. 

The behavior on unsupported browsers generally seemed worse. Instead of error messages in the demo app console, there were just "syntax errors" in browser console. The behavior in Edge 18 also changed from an error message to the "set passcode" button never doing anything. These behaviors may not be worth chasing down. Perhaps our WASM capability detection got broken? 

| Platform | Browser | Notes | Worked? |
| ----------- | ----------- | ------- | ------------ |
| Mac/Big Sur | Chrome 85 | | Yes |
| Mac/Sierra | Safari 14 |  | Yes |
| Mac/Sierra | Safari 10.1 | UI comes up. Initialize fails with "No WASM support"  | No |
| Mac/High Sierra | Safari 11.1 | | Yes* - could not create any documents with the demo app |
| Win 10 | Edge 18 | Set passcode never works. No error? | No |
| Win 10 | Edge 91 |  | Yes  |
| Win 10 | IE 11 | Blank page. Console: "main.js syntax error" | No |
| Win 10 | Opera 76 | | Yes |
| Win 10 | Yandex.12 | No UI. Console: "main.js syntax error"| No |
| iOS/iPhone 12 | Safari | | Yes |
| Android/S8 | Android Browser | Had trouble getting device to boot. Acting weird | No* |
| Android/S9 | Chrome |  | Yes |

clintfred (approved):
Changes seem reasonable.

PR #117, opened 2022-06-07
Update ApiState with new key after changePasscode

### LINKED ISSUE #116: `changePasscode` doesn't change localstorage private key
We ran into this in one of our implementations.

If `changePasscode` succeeds, ironweb doesn't change the `ApiState.encryptedPrivateUserKey` to the new value. This means that any subsequent calls to `changePasscode` will fail until the device is `deauthorized` and re-`initialize`d.

skeet70:
Fixes #116


coltfred (approved):
:+1: LGTM

PR #119, opened 2022-06-07
Fix broken snapshot

skeet70:


.github/scripts/snapshot.js
```diff
@@ -58,9 +58,10 @@ shell.echo(`\n\nBuild successful, version is now set to ${updatedVersion}`);
 
 //Setup git user to our CI user
 shell.echo("\n\nSetting up git so Leeroy can commit our version bump back to main\n");
-shell.exec('git config --global user.email "leeroytravis@ironcorelabs.com"');
+shell.exec('git config --global user.email "ops@ironcorelabs.com"');
 shell.exec('git config --global user.name "Leeroy Travis"');
-shell.exec(`git remote add release "https://${process.env.GIT_ACCESS_TOKEN}@github.com/IronCoreLabs/${process.env.GIT_PROJECT_NAME}.git"`);
+shell.exec(`git remote add release "https://leeroy-jenkins:${process.env.WORKFLOW_PAT}@github.com/IronCoreLabs/${process.env.GIT_PROJECT_NAME}.git"`);
```
cjyar, line 63: I don't think you need the username when you're using a PAT. But if you've tested it, go with your test results over my opinion.

cjyar (approved):

.github/scripts/snapshot.js
skeet70: I tested by cloning locally with this format and it both worked and didn't blow up. Maybe that's because my git configuration normally works, but it at least didn't blow up, so it seems likely it'll work.

I have no idea if the top level config works to get it to use the PAT with later commands but I'm hoping it does!

PR #120, opened 2022-06-07
Simplify the snapshot workflow.

cjyar:
This is modeled after our test case in https://github.com/IronCoreLabs/test1/pull/2.


skeet70 (approved):

coltfred (approved):
:man_shrugging:

PR #130, opened 2022-09-12
Add device delete by id

skeet70:
This also deprecates `deauthorizeDevice`, since `deleteDevice` with no device id argument accomplishes the same goal.

### DIFF
--- a/src/shim/sdk/UserSDK.ts
+++ b/src/shim/sdk/UserSDK.ts
@@ -45,7 +45,6 @@ export const deauthorizeDevice = () => deleteDevice().then((deletedDevice) => ({
  * This method should usually be called whenever the current user logs out of your application or you're aware a device of their's shouldn't have access.
  * @param {number | undefined} deviceId The device id to delete. If undefined, the current device will be deleted and local storage will be cleared.
  */
-// TODO: detect if deviceId is the same as the currently operating device and set deletingCurrentDevice to true
 export const deleteDevice = (deviceId?: number) => {
     checkSDKInitialized();
     const deletingCurrentDevice = deviceId === undefined;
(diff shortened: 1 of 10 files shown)


BobWall23 (approved):

src/shim/sdk/UserSDK.ts
```diff
@@ -34,30 +34,44 @@ export function rotateMasterKey(passcode: string) {
...
+// TODO: detect if deviceId is the same as the currently operating device and set deletingCurrentDevice to true
+export const deleteDevice = (deviceId?: number) => {
     checkSDKInitialized();
-    const payload: MT.DeauthorizeDevice = {
-        type: "DEAUTHORIZE_DEVICE",
-        message: null,
+    const deletingCurrentDevice = deviceId === undefined;
+    const payload: MT.DeleteDevice = {
+        type: "DELETE_DEVICE",
+        message: deletingCurrentDevice ? undefined : deviceId,
```
giarc3, line 54: I think this is simpler?
```suggestion
        message: deviceId,
```

src/shim/sdk/UserSDK.ts
```diff
@@ -34,30 +34,44 @@ export function rotateMasterKey(passcode: string) {
...
+    if (deletingCurrentDevice) {
+        clearParentWindowSymmetricKey();
+    }
+    return FrameMediator.sendMessage<MT.DeleteDeviceResponse>(payload)
         .map(({message}) => {
-            clearSDKInitialized();
-            return {transformKeyDeleted: message};
+            if (deletingCurrentDevice) {
+                clearSDKInitialized();
+            }
```
giarc3, line 65: Could this just be in the above after clearing the window key?

giarc3 (approved):
I couldn't think of a good solution for the TODO

src/shim/sdk/UserSDK.ts
skeet70: It is if we can't come up with a solution to the TODO.

src/shim/sdk/UserSDK.ts
skeet70: No, because we need the SDK initialized still to make the delete request that we're sending to the frame here.

PR #132, opened 2022-09-14
Add delete device by signing key

skeet70:
Peek 2022-09-14 16-43

### DIFF
--- a/src/shim/sdk/UserSDK.ts
+++ b/src/shim/sdk/UserSDK.ts
@@ -68,6 +68,19 @@ export const deleteDevice = (deviceId?: number) => {
 };
 
 /**
+ * Deletes a device by its public signing key.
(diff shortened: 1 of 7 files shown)


src/shim/sdk/UserSDK.ts
```diff
@@ -67,6 +67,19 @@ export const deleteDevice = (deviceId?: number) => {
         .toPromise();
 };
 
+/**
+ * Deletes a device by its public signing key.
+ * @param {number | undefined} deviceId The device id to delete. If undefined, the current device will be deleted and local storage will be cleared.
```
giarc3, line 72: Looks like this was copied from deleteDevice and still needs changing

giarc3 (changes requested):

giarc3 (approved):

PR #134, opened 2022-09-20
Add deleteDeviceByPublicSigningKeyWithJwt

skeet70:
image

This allows someone with access to a jwt for a user to delete a device without logging in and creating another one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The IronWeb SDK NPM releases follow standard [Semantic Versioning](https://semve
 
 **Note:** The patch versions of the IronWeb SDK will not be sequential and might jump by multiple numbers between sequential releases.
 
+## v4.2.1
+- added `deleteDeviceByPublicSigningKeyWithJwt` to the User SDK. This allows someone with access to a JWT for a user to delete a device for that user without initializing the SDK (and creating a new browser device).
+
 ## v4.1.1
 - added `listDevices()` to the User SDK. This lists all devices for the currently logged in user.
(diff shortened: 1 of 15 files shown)


src/frame/sdk/UserApi.ts
```diff
@@ -88,6 +88,12 @@ export const deleteDevice = (deviceId?: number) => {
 export const deleteDeviceBySigningKey = (publicSigningKey: Base64String) =>
     UserApiEndpoints.callUserDeviceDeleteBySigningKey(publicSigningKey).map((r) => r.id);
 
+/**
+* Delete a device from the DB given the public signing key of the device. Uses JWT auth.
+*/
+export const deleteDeviceBySigningKeyWithJwt = (jwtToken: string, publicSigningKey: Base64String) =>
```
coltfred, line 94: It seems like this makes more sense out on the parent IronWeb next to `createNewDeviceKeys`. I couldn't find anything that exists on user that doesn't require the user to be logged in.

If we move it out we also can rename to `deleteDeviceBySigningKey`

coltfred (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The IronWeb SDK NPM releases follow standard [Semantic Versioning](https://semve
 
 **Note:** The patch versions of the IronWeb SDK will not be sequential and might jump by multiple numbers between sequential releases.
 
+## v4.2.1
+- added `deleteDeviceByPublicSigningKeyWithJwt` to the SDK. This allows someone with access to a JWT for a user to delete a device for that user without initializing the SDK (and creating a new browser device).
```
coltfred, line 8: Should update this to the be the new name.

coltfred (approved):

PR #139, opened 2022-12-07
Makes framed `setItem` work in privacy browsers

skeet70:
The feature is not widely supported, so we also do work to make sure the behaviour is the same as it is now if the browser doesn't support `requestStorageAccess`.

I've tested this as much as I can locally and have unit tests that _should_ cover it, but I can't really test it until it's in staging and I can have separate domains for the frame server and the app server.

https://developer.mozilla.org/en-US/docs/Web/Privacy/Storage_Access_Policy/Errors/CookiePartitionedForeign#what_went_wrong
https://developer.mozilla.org/en-US/docs/Web/API/Document/requestStorageAccess
https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning#disable_dynamic_state_partitioning
https://developer.mozilla.org/en-US/docs/Web/Privacy/State_Partitioning#storage_access_api

### DIFF
--- a/src/frame/FrameUtils.ts
+++ b/src/frame/FrameUtils.ts
@@ -15,6 +15,13 @@ function generateFrameStorageKey(userID: string, segmentID: number) {
     return `${ENCRYPTED_DEVICE_KEY_LOCAL_STORAGE_VERSION}-${segmentID}:${userID}-icldaspkn`;
 }
 
+// Helper function to turn `document.requestStorageAccess()` into something more usable.
+const requestStorageAccess = (): Future<never, boolean> =>
+    Future.tryP(() => document.requestStorageAccess())
+        .map(() => true)
+        // since we're catching all errors and converting infallibly to bool, fix the type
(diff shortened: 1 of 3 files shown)


skeet70:
Tested in our SaaS Shield Config Broker in staging vs current bug in prod. Confirmed that after clearing cookies and storage for `ironcorelabs.com`:

1) prod login would show an infinitely spinning beach ball, because it never saved the device keys
2) stage login worked, and once it requested/got the permission for `ironcorelabs.com`, prod started working too (even though it doesn't have the request logic it now had permission).

Reproduced the same behaviour by clearing multiple times and checking it worked the same way. Working on getting confirmation in Safari where prompting on request is more strict.

skeet70:
Tested in Safari too. Can reproduce the bug behavior in Private Browsing, and this deployment to staging fixes it. We couldn't get Safari to ever actually prompt the user though, it always allowed it when requested without asking. Still fixes the bug!

PR #160, opened 2024-06-11
Don't register the message handler with null

skeet70:
This fixes an edge case where the worker is still loading but messages sent to it are accepted and sent to a `null` callback handler.

On the worker side we make sure that there is never a point where the worker registers a `null` function as its handler for `message`. We also add a single `postMessage("ready")` to the parent once we reach the end of the worker file.

On the frame's worker mediator side, we add `Promise` that is only completed when it receives `"ready"` from its worker. Once completed it stays completed for the life of the `WorkerMediator` instance. Only once that internal `Promise` is completed does the worker mediator start sending messages to the worker. 

The combination of these two things makes sure we aren't sending messages before the worker is ready. As far as I can tell from reading the spec and all documentation, none of this should be necessary as the `out` `MessagePort` on the frame side should exist as soon as the worker is called for, and simply buffer messages until a `message` handler is registered on the worker side. In implementation though that didn't hold true.


BobWall23 (approved):

coltfred (approved):
This seems reasonable. I think we can merge it and do further testing in staging.

Should we bump the minor?

PR #171, opened 2025-02-18
Upgrade nightwatch for serializer-javascript

skeet70:
Can't run them locally without correct headless chrome setup


skeet70:
This was for https://github.com/IronCoreLabs/ironweb/security/dependabot/85, but since it's only in testing infrastructure it's not a vuln we care enough about to release a version bump for.

PR #194, opened 2026-02-11
Fix local key encryption issue

coltfred:
When keys are stored in local storage we encrypt them. This was to split the trust between the hosting domain and the domain that hosts our transform service. There was an issue where the IV for encrypting the signing key was re-used for encrypting the device private key.

This fixes that issue by re-encrypting them using separate IVs and generating them correctly for all new devices.

Unfortunately this PR also contains many reformatting changes since prettier hadn't been run on many of the files. If this is a problem I can try to revert some of it. I'd prefer to just get it all fixed up though.

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -81,7 +81,8 @@ interface EncryptedLocalKeys {
     encryptedDeviceKey: Uint8Array;
     encryptedSigningKey: Uint8Array;
     symmetricKey: Uint8Array;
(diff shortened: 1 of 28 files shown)


BobWall23 (approved):

coltfred:
Tested locally on existing keys and new ones. Merging so we can get it into identity and test it in staging.

PR #232, opened 2026-09-18
Remove runtime frame replacement

skeet70:
The runtime frame replacement allowed for a host compromised via XSS or prototype pollution to swap out the frame being communicated with for a malicious one.

This makes the frame address hardcoded in the SDK and closes those attacks. Bundlers can still swap the module that provides the frame address out for testing if needed at build time.


BobWall23 (approved):

coltfred (approved):
I think you should bump the version before merging, but other than that it looks good. Also be sure that #233 merges before the version bump so it's not part of 4.x.

